PR #1, opened 2017-02-24
A helper function to extract Web Interface/API from an object graph

jingt06:


lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+    var meta = og.lookupMetaData(key, id);
+    if (Object.keys(meta).length === 0) {
+      // If meta contains no information, it means this property
+      // is inherited from its prototype. Inherited properties
+      // will be added to thier prototype interface.
+      return false;
+    }
+    if (!showConst) {
+      // If showConst is not turned on, filter the result if
+      // string it is premetive type and writable is false.
```
mdittmer: nit: spelling: primitive

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+function isPosotiveInt(str) {
+  return /^\+?(0|[1-9]\d*)$/.test(str);
+}
+
+// Merge all given arrays into arr, no dupulicate allowed,
+// blacklisted properties are filtered.
+function arrayMerge(arr, opt) {
+  opt = opt || {};
+
+  for (var i = 1; i < arguments.length; i += 1) {
```
mdittmer: nit: here and elsewhere: prefer i++

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+      // string it is premetive type and writable is false.
+      return constType.indexOf(og.getType(keyId)) === -1 || meta.writable !== 0;
+    }
+    return true;
+  });
+  return keys;
+}
+
+// Remove all +, $ chars in given string, return the new string.
+function cleanUp(str) {
```
mdittmer: Maybe add a `TODO` here. It looks like this is doing things like turning `+constructor+` back into `constructor` during filtering and analysis. If there are ever legitimate keys that happen to contain `+` or `$`, then we could run into trouble. It would be more precise to only "clean up" keys that we know were transformed in constructing the object graph.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+function arrayMerge(arr, opt) {
+  opt = opt || {};
+
+  for (var i = 1; i < arguments.length; i += 1) {
+    for (var j = 0; j < arguments[i].length; j += 1) {
+      var str = cleanUp(arguments[i][j]);
+      if (arr.indexOf(str) === -1 &&
+        ((blacklistObjectAttr.indexOf(str) === -1 &&
+        blacklistFunctionAttr.indexOf(str) === -1) || noise) &&
+        !isPosotiveInt(str)) {
```
mdittmer: Is `isPositiveInt` a proxy for "this looks like it's an object graph ID that has been turned into a string"? If so (and we don't use it in other ways), we should give `isPositiveInt` a name that describes how it's used instead of the details of how it works.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+  functionId = og.lookup('prototype', FunctionId);
+  var functionProp = og.getObjectKeys(functionId);
+  for (var i = 0; i < functionProp.length; i += 1) {
+    blacklistFunctionAttr.push(cleanUp(functionProp[i]));
+  }
+}
+
+// Get list of property from Object.prototype
+// and push to blacklistObjectAttr;
+function getObjectBlacklist(og) {
```
mdittmer: Is this the same algorithm as `getFunctionBlacklist`? Maybe we could have a common `getPrototypePropertyProperties` that looks up the property named `prototype` on some given lookup-name or id, then returns an array of its properties.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+
+// Search object id's prototype chain to find the
+// class it belongs to. The class is the prototype
+// with a contructor property.
+function getClassName(id, og) {
+  var ctorId = og.lookup('+constructor+', id);
+  if (ctorId) {
+    return og.getFunctionName(ctorId);
+  }
+  return og.getFunctionName(og.getPrototype(id));
```
mdittmer: Are there platforms where this can blow up the stack? I noticed that there's no guard against `id` being the `null` id.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+
+// Remove blacklist properties and add Function and Object's APIs.
+function postProcess(apiCatalogs, og) {
+  for (let i = 0; i < blacklistProperties.length; i += 1) {
+    var prop = blacklistProperties[i];
+    if (apiCatalogs[prop]) {
+      delete apiCatalogs[prop];
+    }
+  }
+  // Add function and obejct blacklisted property
```
mdittmer: nit: spelling: object

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+  }
+  apiCatalogs[og.getFunctionName(ObjectId)] =
+    apiCatalogs[og.getFunctionName(ObjectId)].concat(blacklistObjectAttr);
+}
+
+/** find API for interface of something with given id.
+  @param {ObjectGraph} og - The obejct graph.
+  @param {JSON} map - the json to store web api for interface
+  @param {int} id - the id currently visiting
+  @param {String} name - the property name from its parent node. (not necessarily interface name)
```
mdittmer: nit: here and elsewhere. Is this a long line? We should be keeping everything within a width of 80.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+  @param {Array} visited - a list of visted properties
+  @param {JSON} opt - know if this object is firstlevel or a __proto__ property.
```
mdittmer: I don't understand the description of this param.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+function extractAPI(og, map, id, name, visited, opt) {
+  opt = opt || {};
+  var firstLevel = opt.firstLevel || false;
+  var proto = opt.proto || false;
+
+  // Return if this object is already visited, except for first level objects.
+  // Still visit it if this is a firstLevel object, since
+  // two different firstlevel object can reference to one object.
+  // We don't want to miss any of them.
+  if (visited.indexOf(id) !== -1 && !firstLevel) return;
```
mdittmer: Visited will get pretty big, no? Should we use a map of `{ <id> : <1 | true | something> }` instead?

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+  // An object/function's Id must be greater than root's Id.
+  // And we don't want to revisit window object.
+  if (id <= rootId) return;
```
mdittmer: This is very efficient, but it's exploiting an implementation detail of `object-graph-js`. Are you trying to return when the `id` indicates a primitive type instead of an object? We should expose an API for that (if one doesn't already exist).

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+  visited.push(id);
+  if (og.isFunction(id) && (proto || firstLevel)) {
+    // This object is a function and it is either a __proto__ or on first level.
+    // Function which is not a __protot__ or first level is not considered as an interface
```
mdittmer: nit: spelling: `__proto__`

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+  @param {ObjectGraph} og - The obejct graph.
+  @param {JSON} map - the json to store web api for interface
+  @param {int} id - the id currently visiting
+  @param {String} name - the property name from its parent node. (not necessarily interface name)
+  @param {Array} visited - a list of visted properties
+  @param {JSON} opt - know if this object is firstlevel or a __proto__ property.
+*/
+function extractAPI(og, map, id, name, visited, opt) {
+  opt = opt || {};
+  var firstLevel = opt.firstLevel || false;
```
mdittmer: What constitutes `firstLevel`? Does it mean this object lives at `window.*` where `* != __proto__`? Does it mean the path we took to get to this object contains no `__proto__` steps?

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const objectGraph = require('object-graph-js').ObjectGraph;
+
+var noise = false;     // True if base property like "caller", "length" will be returned.
```
mdittmer: I'm fine with `let/const` style or `var` style, but we should pick one and use it consistently throughout the project.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+  visited.push(id);
+  if (og.isFunction(id) && (proto || firstLevel)) {
+    // This object is a function and it is either a __proto__ or on first level.
+    // Function which is not a __protot__ or first level is not considered as an interface
+    // eg: chrome.webstore, chrome.runtime.
+    if (id === functionId) return; // Skip if function is Function.prototype;
+    let prototypeId = og.lookup('prototype', id);
+    let ownProperties = getObjectProperties(og, id);
+    let prototypeProps = [];
+    // If function exists as an first level interface, use its property name
```
mdittmer: nit: spelling: `as a first level` (not `an`)

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+    // under window object as interface name, otherwise, use function name.
```
mdittmer: Can you add a comment that gives an example of each to illustrate why both cases need to be covered?

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+    let interfaceName = firstLevel ? name : og.getFunctionName(id);
+    if (prototypeId) prototypeProps = getObjectProperties(og, prototypeId);
+    arrayMerge(ownProperties, prototypeProps);
+    let meaningfulProps = setMinus(ownProperties, blacklistFunctionAttr);
+    if (meaningfulProps.length > 0) {
+      // If this function and its prototype contains meaning ful properties.
```
mdittmer: nit: spelling: no space in `meaningful`

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+      if (!map[interfaceName]) map[interfaceName] = [];
+      arrayMerge(map[interfaceName], getObjectProperties(og, id),
```
mdittmer: This looks like we curate a collection of meaningful properties, then, if there are any, we store all properties, meaningful or not. Is that intentional?

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+          getObjectProperties(og, prototypeId));
+    }
+    if (og.getPrototype(id)) {
+      extractAPI(og, map, og.getPrototype(id), `${name}__proto__`, visited, {proto: true});
```
mdittmer: Looks like you're using `undefined`-is-falsey in your `opt.<option>` checks. That should be documented in the function header.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+    }
+  } else if (!og.isFunction(id)) {
+    // Object is not a function.
```
mdittmer: Rather than stating the if condition (which is already pretty readable), could you state the complete condition for the `else if` branch? `// Object is not <original if condition>, nor is it a function`, or something similar.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+    let ownProperty = getObjectProperties(og, id);
+    if (setMinus(ownProperty, blacklistObjectAttr).length > 0) {
+      let className = getClassName(id, og);
+      if ((!className || className === "Object") && firstLevel) {
+        // If object is at first level and its class name is "Object",
+        // this object is considered as an object library.
```
mdittmer: Can you give an example to clarify what's meant by "object library"?

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+        let prototypeId = og.lookup('prototype', id);
+        // in Firefox Edge ..., objects can have prototype properties.
```
mdittmer: What's with the "...", are you unsure if others do this as well? If so, it's clearer to say something like `In Firefox, Edge, (and maybe other vendors?), objects <...>`.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+        let prototypeProps = [];
+        if (prototypeId) {
+          prototypeProps = getObjectProperties(og, prototypeId);
+        }
+        // if (!map[name]) map[name] = [INTERFACE];
```
mdittmer: Just commented-out from previous implementation? If so, we should just delete this line. If it's still here because sometimes you uncomment it for debugging, please add a comment above explaining how it's used for debugging.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+        if (!map[name]) map[name] = [];
+        arrayMerge(map[name], getObjectProperties(og, id), prototypeProps);
+      } else {
+        // This is not a object library, try to add its property to it's prototype.
```
mdittmer: I'm having trouble following how this else-case is used. Could you add an example, and possibly a brief description of why the example fails the if, fails the else-if, and then this work still needs to be done?

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+        arrayMerge(map[className], getObjectProperties(og, id));
+      }
+    }
+    if (og.getPrototype(id)) {
+      // Visit object __proto__ if its __proto__ exists.
+      extractAPI(og, map, og.getPrototype(id), `${name}__proto__`, visited, {proto: true});
+    }
+  }
+
+  // Dont visit __proto__'s properties.
```
mdittmer: nit: spelling: `Don't`

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+
+var blacklistFunctionAttr = ['prototype'];
+// blacklistFunctionAttr is list of property of Function.prototype,
+// prototype is not a part of function, but we also wish to blacklist it.
+
+var blacklistObjectAttr = [];
+// blacklistObjectAttr is list of property of Object.prototype.
+
+var constType = ["boolean", "number", "string"];
+// Primitive types which is constant will be blacklisted.
```
jingt06: Constant properties are blacklisted. Constant properties are defined as: it is primitive type and its writable equals to 0.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+
+var rootId, ObjectId, FunctionId, objectId, functionId;
+
+// A wrapper function of og.getObjectKeys, constant will
+// be filtered if showConst is not turned on.
+function getObjectProperties(og, id) {
+  var keys = og.getObjectKeys(id);
```
jingt06: This function may return non-own keys. We have stored inherited properties in object graph, since they are not own properties of an object, they do not have metadata.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -0,0 +1,279 @@
...
+var showConst = false; // True if constant number/string/boolean value will be returned.
+
+var blacklistProperties = [
+  'toString',    // Do not visit object's toString.
+  'constructor',  // Do not visit constructor functions, they have prototype but not interface.
+  'CSS2Properties' // These are known bugs in Firefox.
+];
+
+var blacklistFunctionAttr = ['prototype'];
+// blacklistFunctionAttr is list of property of Function.prototype,
```
jingt06: For any function interface, blacklist properties include Object.getOwnPropertyNames(Function.prototype) and "prototype"

lib/web_catalog/api_extractor.es6.js
jingt06: The purpose of isPosotiveInt is to filter non-meaningful array's properties. For example, Object.getOwnPropertyNames(document.scripts) will return ["0", "1", "2" ... ].

lib/web_catalog/api_extractor.es6.js
jingt06: Two first level object may reference to the same object. For example, window.webkitMediaStream and window.MediaStream are same function. If only use function name, webkitMediaStream interface will not be captured.

lib/web_catalog/api_extractor.es6.js
jingt06: Object library is something like Math. It is an instance of Object. If something is not a object library, it can be some class's instance. For example, document.body.style contains meaningful properties of CSSStyleDeclaration, then its properties will be added to CSSStyleDeclaration interface.

lib/web_catalog/api_extractor.es6.js
mdittmer: I see. Adding a comment would help. We should be sure to mention this in the "Improving API Confluence" memo section about filtering too.

PR #12, opened 2017-03-07
set up for travis CI

jingt06:


jingt06:
Done.

.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: node_js
+node_js:
+ - "6"
+addons:
+  firefox: latest
+before_script:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
```
mdittmer: Should we be adding a `sleep` statement here to avoid test flake due to xvfb not starting quickly enough? https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI

config/karma.conf.js
```diff
@@ -95,6 +96,7 @@ function configurator(config) {
     // Concurrency level
     // how many browser should be started simultaneous
     concurrency: Infinity,
+
```
mdittmer: nit: Don't need this extra line.

PR #13, opened 2017-03-08
FOAM models for browser, web interface and browser APIs

jingt06:
- FOAM models for browser, web interface and browsers added.
- Essential methods for webAPIs are added.
- Change webpack-helper.js to hardcoded require(), since require.context does not give any dependency options.

### DIFF
--- a/lib/web_apis/browser.es6.js
+++ b/lib/web_apis/browser.es6.js
@@ -1,0 +1,66 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+foam.CLASS({
+  name: 'Browser',
+  package: 'com.web.api',
+  properties: [
+    {
+      name: 'browserName',
+      required: true,
+    },
+    {
+      name: 'browserVersion',
+      required: true,
+    },
+    {
+      name: 'OS',
+      required: true,
+    },
+    {
+      name: 'OSVersion',
+      required: true,
+    },
+    {
+      name: 'id',
+    },
+  ],
+  methods: [
+    {
+      name: 'init',
+      document: `After the browser object is created, hashCode function is
+        used to assign this browser a unique browser id`,
+      code: function() {
+        this.id = foam.util.hashCode(this);
+      },
+    },
+    {
+      name: 'getBrowserKey',
+      documentation: `Get the key of this browser, including browser name,
+        browser version, OS and OS version.`,
+      returns: {
+        documentation: `The key of this browser.`,
+        typeName: `String`,
+      },
+      code: function() {
+        return [this.browserName, this.browserVersion, this.OS,
+          this.OSVersion].join('_');
+      },
+    },
+  ],
+});
--- a/lib/web_apis/browser_api.es6.js
+++ b/lib/web_apis/browser_api.es6.js
@@ -1,0 +1,27 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+foam.CLASS({
+  name: 'BrowserAPI',
+  package: 'com.web.api',
+  properties: [
+    'id',
+    'browserId',
+    'interfaceId',
+  ],
+});
--- a/lib/web_apis/web_apis.es6.js
+++ b/lib/web_apis/web_apis.es6.js
@@ -1,0 +1,348 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+require('./browser.es6');
+require('./browser_api.es6');
+require('./web_interface.es6');
+
+foam.CLASS({
+  name: 'WebAPIs',
+  package: 'com.web.api',
+  requires: [
+    'com.web.api.Browser',
+    'com.web.api.BrowserAPI',
+    'com.web.api.WebInterface',
+    'foam.dao.EasyDAO',
+    'foam.mlang.sink.GroupBy',
+    'foam.dao.ArrayDAO',
+    'foam.mlang.ExpressionsSingleton',
+  ],
+  properties: [
+    {
+      name: 'browserAPIs',
+      documentation: `A DAO that contains pairs of browser Id
+        and interface Id.`,
+      factory: function() {
+        return this.EasyDAO.create({
+          name: 'BrowserAPIDAO',
+          of: this.BrowserAPI,
+          daoType: 'MDAO',
+        });
+      },
+    },
+    {
+      name: 'browsers',
+      documentation: `A DAO that contains browser names, versions
+        OS, OS versions and their browser Id.`,
+      factory: function() {
+        return this.EasyDAO.create({
+          name: 'BrowserDAO',
+          of: this.Browser,
+          daoType: 'MDAO',
+        });
+      },
+    },
+    {
+      name: 'interfaces',
+      documentation: `A DAO that contains interface names, API names,
+        and their interfaceId`,
+      factory: function() {
+        return this.EasyDAO.create({
+          name: 'InterfaceDAO',
+          of: this.WebInterface,
+          daoType: 'MDAO',
+        });
+      },
+    },
+    {
+      name: 'mlang',
+      factory: function() {
+        return this.ExpressionsSingleton.create();
+      },
+    },
+  ],
+  methods: [
+    {
+      name: 'importAPI',
+      documentation: 'Import interface/API for a given version of browser.',
+      args: [
+        {
+          name: 'browserName',
+          documentation: `The name of imported browser.`,
+          typeName: 'String',
+        },
+        {
+          name: 'browserVersion',
+          documentation: `The version of imported browser.`,
+          typeName: 'String',
+        },
+        {
+          name: 'OSName',
+          documentation: `The name of operating system.`,
+          typeName: 'String',
+        },
+        {
+          name: 'OSVersion',
+          documentation: `The version of operating system.`,
+          typeName: 'String',
+        },
+        {
+          name: 'apiCatalog',
+          documentation: `The web catalog of this version of
+            browser, tested on the given operating system and version.
+            The json object should be the JSON object returned from
+            com.web.catalog.apiExtractor`,
+          typeName: 'JSON',
+        },
+      ],
+      code: function(browserName, browserVersion,
+        OSName, OSVersion, apiCatalog) {
+        let browser = this.Browser.create({
+          browserName: browserName,
+          browserVersion: browserVersion,
+          OS: OSName,
+          OSVersion: OSVersion,
+        });
+        let browserId = browser.id;
+        // Check?
+        this.browsers.put(browser);
+        let interfaceNames = Object.keys(apiCatalog);
+        for (let i = 0; i < interfaceNames.length; i += 1) {
+          let interfaceName = interfaceNames[i];
+          for (let j = 0; j < apiCatalog[interfaceName].length; j += 1) {
+            let APIName = apiCatalog[interfaceName][j];
+            let webInterface = this.WebInterface.create({
+              interfaceName,
+              APIName,
+            });
+            let interfaceId = webInterface.id;
+            this.interfaces.put(webInterface);
+            this.browserAPIs.put(this.BrowserAPI.create({
+              browserId,
+              interfaceId,
+              id: `${browserId}.${interfaceId}`,
+            }));
+          }
+        }
+      },
+    },
+    {
+      name: 'getBrowserKeys',
+      documentation: `Asynch function to return an array of stored
+        browsers' key. browserName and version are optional arguments
+        to filter keys.`,
+      args: [
+        {
+          name: 'browserName',
+          documentation: `The optional argument that the returned browser
+            key will match the given name`,
+          typeName: 'String',
+        },
+        {
+          name: 'version',
+          documentation: `The optional argument that the returned browser
+            key will match the given version`,
+          typeName: 'String',
+        },
+      ],
+      returns: {
+        typeName: 'StringArray',
+        documentation: `Return a string array contains matching browser keys`,
+      },
+      code: function(browserName, version) {
+        return new Promise((resolve) => {
+          this.browsers.select().then((arrayDAO) => {
+            resolve(arrayDAO.a
+            .map((browser) => browser.getBrowserKey())
+            .filter((key) => {
+              if (browserName &&
+              key.split('_')[0].indexOf(browserName) !== 0) {
+                return false;
+              }
+              if (version &&
+              key.split('_')[1].indexOf(version) !== 0) {
+                return false;
+              }
+              return true;
+            }));
+          });
+        });
+      },
+    },
+    {
+      name: 'getBrowserId_',
+      args: [
+        {
+          name: 'browserKey',
+          typeName: 'String',
+          documentation: `The string must be a key of browser`,
+        },
+      ],
+      returns: {
+        typeName: 'Number',
+        documentation: `The id of this browser`,
+      },
+      code: function(browserKey) {
+        let browserArr = browserKey.split('_');
+        return this.Browser.create({
+          browserName: browserArr[0],
+          browserVersion: browserArr[1],
+          OS: browserArr[2],
+          OSVersion: browserArr[3],
+        }).id;
+      },
+    },
+    {
+      name: 'toMap_',
+      documentation: `Takes an array of browserAPIs and an array of
+        browser keys, produce a JSON containing browser and interface
+        relationship`,
+      args: [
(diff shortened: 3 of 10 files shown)


lib/web_apis/browser_api.es6.js
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+foam.CLASS({
+  name: 'BrowserAPI',
+  package: 'com.web.api',
```
mdittmer: Please add class-level `documentation: ...`.

lib/web_apis/browser.es6.js
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+foam.CLASS({
+  name: 'Browser',
+  package: 'com.web.api',
```
mdittmer: Please add class-level `documentation: ...`.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+ */
+'use strict';
+
+require('./browser.es6');
+require('./browser_api.es6');
+require('./web_interface.es6');
+
+foam.CLASS({
+  name: 'WebAPIs',
+  package: 'com.web.api',
```
mdittmer: Please add class-level `documentation: ...`.

lib/web_apis/browser_api.es6.js
```diff
@@ -0,0 +1,27 @@
...
+  properties: [
+    'id',
+    'browserId',
```
mdittmer: We should model relationships over these ids.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+    'foam.dao.ArrayDAO',
+    'foam.mlang.ExpressionsSingleton',
+  ],
+  properties: [
+    {
+      name: 'browserAPIs',
+      documentation: `A DAO that contains pairs of browser Id
+        and interface Id.`,
+      factory: function() {
+        return this.EasyDAO.create({
```
mdittmer: We should specify indexing over all ids here.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+    'com.web.api.WebInterface',
+    'foam.dao.EasyDAO',
+    'foam.mlang.sink.GroupBy',
+    'foam.dao.ArrayDAO',
+    'foam.mlang.ExpressionsSingleton',
+  ],
+  properties: [
+    {
+      name: 'browserAPIs',
+      documentation: `A DAO that contains pairs of browser Id
```
mdittmer: Here and other DAO-containg properties: Please document the context rather than the contents of the type (which is clear from the `of: ...` on the DAO). Is this all BrowserAPIs this WebAPIs component knows about? Is it somehow filtered? Is it lazily loaded or all dumped into the component at once?

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+        and interface Id.`,
+      factory: function() {
+        return this.EasyDAO.create({
+          name: 'BrowserAPIDAO',
```
mdittmer: nit: When I have back-to-back acroynms, I tend to `CamelCase` the first one so that they don't "bleed into each other" for the reader. Here I'd use `BrowserApiDAO`.

lib/web_apis/browser.es6.js
```diff
@@ -0,0 +1,66 @@
...
+    {
+      name: 'browserName',
+      required: true,
+    },
+    {
+      name: 'browserVersion',
+      required: true,
+    },
+    {
+      name: 'OS',
```
mdittmer: nit: here and `OSVersion`; don't start property names with a capital, even if it's an acronym. Prefer `os` and `osVersion`.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+          daoType: 'MDAO',
+        });
+      },
+    },
+    {
+      name: 'browsers',
+      documentation: `A DAO that contains browser names, versions
+        OS, OS versions and their browser Id.`,
+      factory: function() {
+        return this.EasyDAO.create({
```
mdittmer: @jacksonic we only need a primary key index here. Will we get it for free, or do we need to specify it?

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+          documentation: `The name of imported browser.`,
+          typeName: 'String',
+        },
+        {
+          name: 'browserVersion',
+          documentation: `The version of imported browser.`,
+          typeName: 'String',
+        },
+        {
+          name: 'OSName',
```
mdittmer: Here and elsewhere: as with property names, don't start with capital unless it's a type. Prefer `osName`, `osVersion`, etc..

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+      code: function(browserName, browserVersion,
+        OSName, OSVersion, apiCatalog) {
+        let browser = this.Browser.create({
+          browserName: browserName,
+          browserVersion: browserVersion,
+          OS: OSName,
+          OSVersion: OSVersion,
+        });
+        let browserId = browser.id;
+        this.browsers.put(browser);
```
mdittmer: What's the context of the call sites for this? If we find the browser is already in the DAO, should we exit early?

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+        let interfaceNames = Object.keys(apiCatalog);
+        for (let i = 0; i < interfaceNames.length; i += 1) {
```
mdittmer: nit: here and elsewhere: prefer `i++`.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+        browser keys, produce a JSON containing browser and interface
+        relationship`,
+      args: [
+        {
+          name: 'browserAPIs',
+          typeName: 'BrowserAPI Array',
+          documentation: `An array returned from DAO.select.`,
+        },
+        {
+          name: 'brwoserKeys',
```
mdittmer: nit: spelling: `browserKeys`

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+            let webInterface = this.WebInterface.create({
+              interfaceName,
+              APIName,
+            });
+            let interfaceId = webInterface.id;
+            this.interfaces.put(webInterface);
+            this.browserAPIs.put(this.BrowserAPI.create({
+              browserId,
+              interfaceId,
+              id: `${browserId}.${interfaceId}`,
```
mdittmer: Let's let the model take care of this. That way if other code ever constructs these we don't risk having multiple patterns for constructing ids by mistake.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+          typeName: 'String',
+        },
+      ],
+      returns: {
+        typeName: 'StringArray',
+        documentation: `Return a string array contains matching browser keys`,
+      },
+      code: function(browserName, version) {
+        return new Promise((resolve) => {
+          this.browsers.select().then((arrayDAO) => {
```
mdittmer: nit: this is an `ArraySink`. Prefer name `arraySink`.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+      },
+    },
+    {
+      name: 'getBrowserKeys',
+      documentation: `Asynch function to return an array of stored
+        browsers' key. browserName and version are optional arguments
+        to filter keys.`,
+      args: [
+        {
+          name: 'browserName',
```
mdittmer: nit: Let's use consistent names. Go with `name`, `version` (contextualized by `getBrowserKeys` clarifies meaning) or `browserName`, `browserVersion` for clarity (and, arguably, verbosity).

lib/web_apis/web_apis.es6.js
mdittmer: We're using a thenable, so I don't think we need `new Promise(...)`. Just `return this.browsers.select().then((arrayDAO) => { ... });`

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+              interfaceId,
+              id: `${browserId}.${interfaceId}`,
+            }));
+          }
+        }
+      },
+    },
+    {
+      name: 'getBrowserKeys',
+      documentation: `Asynch function to return an array of stored
```
mdittmer: I'm not sure we need this, do we? `myBrowserDAO.where(AND(EQ(Browser.BROWSER_NAME, 'name'), EQ(Browser.BROWSER_VERSION, 'version')))` gives us this.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+                return false;
+              }
+              return true;
+            }));
+          });
+        });
+      },
+    },
+    {
+      name: 'getBrowserId_',
```
mdittmer: As above regarding need for this. We can query our dao for the `Browser` object and ask it for its key. Are you thinking that making such a query be synchronous is critical?

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+              }));
+          }
+          Promise.all(promises).then(function() {
+            resolve(apiMap);
+          });
+        }.bind(this));
+      },
+    },
+    {
+      name: 'toMap',
```
mdittmer: Let's discuss this and its helper offline. I don't think we've been over this design detail.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+          documentation: `A string array contains valid browser keys`,
+        },
+      ],
+      returns: {
+        typeName: 'String',
+        documentation: `Returns a string of csv format`,
+      },
+      code: function(browserKeys) {
+        return new Promise(function(resolve) {
+          this.toMap(browserKeys).then(function(result) {
```
mdittmer: Don't need `new Promise(...)`, just return from thenable.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+          typeName: 'StringArray',
+          documentation: `A string array contains valid browser keys`,
+        },
+      ],
+      returns: {
+        typeName: 'String',
+        documentation: `Returns a string of csv format`,
+      },
+      code: function(browserKeys) {
+        return new Promise(function(resolve) {
```
mdittmer: Let's discuss this strategy. What I had in mind after our meeting was that procedures such as this would use something like:

```
// DAO query pseudo-code:
SELECT interface_id FROM apis ORDER BY interface_name
SELECT * FROM browser_apis ORDER BY browser_id GROUP BY interface_id
for each interface_id: use GROUP BY result to extract browser_ids that are True; fill in True/False in ordered-by-browser_id columns
```

Maybe there are results from experimenting with different approaches that led to this implementation. Let's chat offline.

lib/web_apis/web_interface.es6.js
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+foam.CLASS({
+  name: 'WebInterface',
+  package: 'com.web.api',
+  properties: [
```
mdittmer: Types & documentation.

lib/web_apis/browser.es6.js
```diff
@@ -0,0 +1,66 @@
...
+      name: 'id',
+    },
+  ],
+  methods: [
+    {
+      name: 'init',
+      document: `After the browser object is created, hashCode function is
+        used to assign this browser a unique browser id`,
+      code: function() {
+        this.id = foam.util.hashCode(this);
```
jingt06: The hashCode function should only be called once in initialization. After that, the browserId can be accessed via browser.id.

mdittmer:
That's true, but it's still a potential robustness issue. For example, I
think this logic will results in `myBrowserObject.clone()` producing a
browser object with a different id, which is unexpected.

lib/web_apis/browser.es6.js
```diff
@@ -0,0 +1,66 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+foam.CLASS({
+  name: 'Browser',
+  package: 'com.web.api',
+  properties: [
```
kgrgreer: "final: true" lets us speed-up property getters in cases where performance is really critical, but you shouldn't use unless you have a good reason

lib/web_apis/browser.es6.js
```diff
@@ -0,0 +1,66 @@
...
+      name: 'OS',
+      required: true,
+    },
+    {
+      name: 'OSVersion',
+      required: true,
+    },
+    {
+      name: 'id',
+    },
```
kgrgreer: No trailing commas.

lib/web_apis/browser.es6.js
kgrgreer: You could call foam.uuid.randomGUID() instead. Also, more FOAM-ish to put this in the factory: for the 'id' property. hashCode() is not guaranteed to be unique.

lib/web_apis/browser.es6.js
```diff
@@ -0,0 +1,66 @@
...
+    {
+      name: 'init',
+      document: `After the browser object is created, hashCode function is
+        used to assign this browser a unique browser id`,
+      code: function() {
+        this.id = foam.util.hashCode(this);
+      },
+    },
+    {
+      name: 'getBrowserKey',
```
kgrgreer: If you're calling this much, you could make it a Property with an expression, in which case you would get caching for free. Not a big deal though if you don't call much.

lib/web_apis/web_apis.es6.js
kgrgreer: That's the Google standard: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

lib/web_apis/web_apis.es6.js
kgrgreer: Primary key index isn't for free, but it is implicit.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+          of: this.WebInterface,
+          daoType: 'MDAO',
+        });
+      },
+    },
+    {
+      name: 'mlang',
+      factory: function() {
+        return this.ExpressionsSingleton.create();
+      },
```
kgrgreer: No trailing commas.

lib/web_apis/web_apis.es6.js
kgrgreer: See foam.dao.GUIDDAO or foam.dao.SequenceNumberDAO for examples.

lib/web_apis/browser.es6.js
mdittmer: I think the preference for `hasCode()` was:
1. In practice, it should be unique for this small data set,
2. It's numeric (which GUIDs are not); this speeds up indexes over it (is this assumption wrong? @jingt06 have you run experiments on this?)

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,347 @@
...
+                {interfaceName: {APIName: [Boolean ...]} ...}`,
+      },
+      code: function(browserKeys) {
+        return new Promise(function(resolve) {
+          let browserIds = browserKeys.map((key) => this.getBrowserId_(key));
+          this.browserAPIs
+            .where(this.mlang.IN(this.BrowserAPI.BROWSER_ID, browserIds))
+            .select()
+            .then((arrayDao) => {
+              resolve(this.toMap_(arrayDao.a, browserKeys));
```
kgrgreer: We have a MAP() mLang which lets you do this as part of the select() call.

lib/web_apis/web_apis.es6.js
mdittmer: "That" being my recommendation? The current name? Or something else? What you linked seems to imply that `DAO` should be `Dao`.

lib/web_apis/web_apis.es6.js
mdittmer: Yes. I misspoke when I said "free". I meant "will be constructed automatically without further configuration". Is that what you mean by "implicit"?

lib/web_apis/web_interface.es6.js
```diff
@@ -0,0 +1,37 @@
...
+    'APIName',
+    'id',
+  ],
+  methods: [
+    {
+      name: 'init',
+      document: `After the WebInterface is created, hashCode function is used
+        to assign this browser a unique browser id`,
+      code: function() {
+        this.id = foam.util.hashCode(this);
```
kgrgreer: Same comments as when this pattern was used above.

lib/web_apis/web_apis.es6.js
mdittmer: My only concern with GUID or seq-no is that it would be nice to have the property that the id be a function of identifying properties. That way different systems operating over the same data construct the same ids.

@kgrgreer Does that seem feasible, or do you anticipate collision issues with `hashCode()` or similar?

lib/web_apis/browser.es6.js
mdittmer: @kgrgreer this is acceptable in Google style and we're not following FOAM style guidelines for this project. I think trailing commas are used consistently in this project. If not, we should pick one and stick with it.

lib/web_apis/web_apis.es6.js
mdittmer: @jingt06 Probably disregard (as mentioned above).

lib/web_apis/browser.es6.js
kgrgreer: Yes, that's a good reason.

lib/web_apis/browser.es6.js
kgrgreer: Would a simple sequence number be sufficient?

jingt06: Using hashCode() let browser with same name. version, OS will have same browser id. Then the client side can get the browser Id without querying the browserDAO, which I think may be more efficient.
Another thing I found is Call GroupBy on numeric properties are much more faster than String properties.

lib/web_apis/web_apis.es6.js
kgrgreer: So, the standard means it should be called BrowserApiDao, but we name DAO's "DAO", since they're special.

lib/web_apis/web_apis.es6.js
kgrgreer: Yes.

lib/web_apis/browser.es6.js
kgrgreer: Okay, then  trail away.

lib/web_apis/browser.es6.js
foolip: I guess we have to use a reverse domain name? In that case, org.chromium plus whatever token is going to be the interesting part of the actual URL. If we're calling the thing web-apis, then org.chromium.web-apis?

mdittmer:
We should avoid using hyphens. We could do `org.chromium.apis` or
`org.chromium.apis.web`.

On Fri, Mar 10, 2017 at 6:36 AM, Philip Jägenstedt <notifications@github.com

PR #17, opened 2017-03-13
FOAM models of BrowserAPI and WebAPI using one table implementation.

jingt06:


lib/web_apis/browser_api.es6.js
```diff
@@ -0,0 +1,86 @@
...
+'use strict';
+
+foam.CLASS({
+  name: 'BrowserAPI',
+  package: 'org.chromium.apis.web',
+  documentation: `BrowserAPI is a FOAM model that contains browser name,
+  	browser version, OS, OS version, interface name and API name.`,
+  properties: [
+    {
+      name: 'browserName',
```
mdittmer: nit: here and elsewhere: `class` first.

lib/web_apis/browser_api.es6.js
```diff
@@ -0,0 +1,86 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+foam.CLASS({
+  name: 'BrowserAPI',
+  package: 'org.chromium.apis.web',
+  documentation: `BrowserAPI is a FOAM model that contains browser name,
```
mdittmer: nit: Mentioning that it's a `FOAM model` is unnecessary; that's clear from the context.

lib/web_apis/browser_api.es6.js
```diff
@@ -0,0 +1,86 @@
...
+  	browser version, OS, OS version, interface name and API name.`,
+  properties: [
```
mdittmer: Please add `documentation: ...` for each property. In particular, `interfaceName` and `apiName` need to be clearly documented as being `<interface-name>(.prototype)?.<api-name>` from Javascript bindings.

lib/web_apis/browser_api.es6.js
```diff
@@ -0,0 +1,86 @@
...
+    {
+      name: 'interfaceKey',
+      documentation: `A key for this pair of interface and API.`,
+      class: 'String',
+      expression: function(interfaceName, apiName) {
+        return`${interfaceName}_${apiName}`;
+      },
+    },
+  ],
+  ids: [
```
mdittmer: nit: `ids` as metadata: put above `properties`, `methods`, etc..

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+          name: 'BrowserAPIDAO',
+          of: this.BrowserAPI,
+          daoType: 'MDAO',
+        }).addPropertyIndex(this.BrowserAPI.BROWSER_KEY)
+        .addPropertyIndex(this.BrowserAPI.INTERFACE_KEY);
+      },
+    },
+    {
+      name: 'mlang',
+      documentation: `FOAM's mlang expressiongs, it supports Math and
```
mdittmer: No need to say it's a FOAM mlang; clear from the context. However, please document the _why_ and the _how_: What's this used for? Is it usually injected or set by some other code later?

lib/web_apis/browser_api.es6.js
```diff
@@ -0,0 +1,86 @@
...
+    },
+    {
+      name: 'apiName',
+      required: true,
+      final: true,
+      class: 'String',
+    },
+    {
+      name: 'browserKey',
+      documentation: `An unique key for this browser.`,
```
mdittmer: Here and `interfaceKey`: These seem superfluous, but you have them for a reason: Avoid the need for `CONCAT` mLang or similar to be able to `groupBy` `browserName, browserVersion, osName, osVersion`. Please explain this in the documentation.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+          documentation: `The web catalog of this version of
+            browser, tested on the given operating system and version.
+            The json object should be the JSON object returned from
+            com.web.catalog.apiExtractor`,
+          typeName: 'JSON',
+        },
+      ],
+      code: function(browserName, browserVersion,
+        osName, osVersion, apiCatalog) {
+        let browserKey = `${browserName}_${browserVersion}_${osName}_${osVersion}`;
```
mdittmer: There should be a shared implementation for constructing these keys. I think we should do one of two things here:

(1) Create a `foam.LIB` that implements `${browserName}_${browserVersion}_${osName}_${osVersion}`. Make the `browserKey` `expression` call out to the lib, and also call out to the lib here.
(2) At this point, construct a `BrowserAPI`, then ask for its `browserKey`.

Both solutions are fine by me. We may create more garbage with (2) though.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+        return this.browserAPIs.select(this.GroupBy.create({
```
mdittmer: Pro tip: Make this `implements: [ 'foam.mlang.Expressions' ]` and use `this.GROUP_BY(this.BrowserAPI.BROWSER_KEY, this.COUNT())`. See `foam2/src/mlang/mlang.js` for details on which the `MLANG_NAME(...)`s you get from the `implements`.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+      ],
+      returns: {
+        typeName: 'JSON',
+        documentation: `Returns a JSON of the form:
+                {interfaceName: {apiName: [Boolean ...]} ...}
+                This structure is easy for displaying a nested table.`,
+      },
+      code: function(browserKeys) {
+        return this.browserAPIs
+          .where(this.mlang.IN(this.BrowserAPI.BROWSER_KEY, browserKeys))
```
mdittmer: I think add the `implements` mentioned above will allow you to do `this.IN`

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+          typeName: 'StringArray',
+          documentation: `A string array contains valid browser keys.`,
+        },
+      ],
+      returns: {
+        typeName: 'String',
+        documentation: `Returns a string of csv format.`,
+      },
+      code: function(browserKeys) {
+        return this.toMap(browserKeys).then(function(result) {
```
mdittmer: Can't we just:

```
this.browserAPIs.where(this.IN(this.BrowserAPI.BROWSER_KEY, browserKeys))
  .select(this.GROUP_BY(this.BrowserAPI.INTERFACE_KEY))
  .then(function() { /* Spit out rows according to missing/not-missing objects in groups */ })
```

?

Of course, we'll need to output the header first, but that's fine.

test/any/web_apis/web-apis-integration.es6.js
```diff
@@ -0,0 +1,107 @@
+/**
```
mdittmer: nit: here and other test files: file name: `web_apis-integration.es6.js` to match `<file-under-test>-integration.es6.js`.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+          name: 'browserKeys',
+          typeName: 'StringArray',
+          documentation: `A list of browser keys that contained in the
+            browserAPIs.`,
+        },
+      ],
+      returns: {
+        typeName: 'JSON',
+        documentation: `Returns a JSON of the form:
+            {interfaceName: {apiName: [Boolean ...]} ...}
```
arobins: Maybe I'm missing something, but the result seems to contain _header as a property? I think I'd rather have this output something (interface->API->browserKey->bool maybe) without the header row, and have the code that cares about the header (CSV output) do the manipulation.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+'use strict';
+
+require('./browser_api.es6');
+
+foam.CLASS({
+  name: 'WebAPIs',
+  package: 'org.chromium.apis.web',
+  documentation: `WebAPIs is the main model of this project. It has a
+    property called browserAPIs which is a FOAM DAO. It contains interface and
+    API information for major browsers and version for recent two years. It
```
jingt06: Our first goal is to analyze confluence metrics for recent 2 years.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+    has methods to transform the data model from Google Cloud Datastore
+    to CSV format String or JSON structure to be displayed in client side.
+    It also has methods to do data analysis.`,
```
jingt06: My intention is to hide model-related code in this FOAM class, so my client side code does not need to worry about the change of data models.

arobins: So why not have this model for storing the raw data and a separate model for the API support matrix? You already have that in the toMap/toCSV methods, just not declared as a class.

lib/web_apis/web_apis.es6.js
jingt06: GROUP_BY is not exposed as a method of mlang.Expressions. [foam2/mlang.js](https://github.com/foam-framework/foam2/blob/master/src/foam/mlang/mlang.js#L1606)

mdittmer: @kgrgreer is "no GROUP_BY in mlang.Expressions" a bug or a feature?

lib/web_apis/browser_api.es6.js
jingt06: Done

lib/web_apis/browser_api.es6.js
jingt06: Done

lib/web_apis/browser_api.es6.js
jingt06: Comment added

lib/web_apis/browser_api.es6.js
jingt06: Done

lib/web_apis/web_apis.es6.js
jingt06: Done. Used groupBy to build apiMatrix.

test/any/web_apis/web-apis-integration.es6.js
jingt06: Done

lib/web_apis/web_apis.es6.js
jingt06: Done, I separated the web_apis class to a client side api_matrix class and a api-importer class.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+    'foam.dao.ArrayDAO',
+    'foam.mlang.ExpressionsSingleton',
+    'foam.mlang.sink.GroupBy',
+    'foam.mlang.sink.Count',
+  ],
+  properties: [
+    {
+      name: 'browserAPIs',
+      documentation: `This should be a server GCD DAO that contains
+        interfaece, API information for major browsers and versions
```
jingt06: Done.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+        for recent 2 years. It is implemented as a MDAO for now.`,
+      factory: function() {
+        return this.EasyDAO.create({
+          name: 'BrowserAPIDAO',
+          of: this.BrowserAPI,
+          daoType: 'MDAO',
+        }).addPropertyIndex(this.BrowserAPI.BROWSER_KEY)
+        .addPropertyIndex(this.BrowserAPI.INTERFACE_KEY);
```
jingt06: Done

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+                apiName,
+              }));
+            }
+          }
+          return true;
+        });
+      },
+    },
+    {
+      name: 'toMap_',
```
jingt06: changed to apiMatrix

lib/web_apis/web_apis.es6.js
jingt06: changed the output to {interfaceName: {apiName: {BrowserKey: true, ...} ...} ... } ...}

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+      args: [
+        {
+          name: 'browserAPIs',
+          typeName: 'BrowserAPI Array',
+          documentation: `An array returned from DAO.select.`,
+        },
+        {
+          name: 'browserKeys',
+          typeName: 'StringArray',
+          documentation: `A list of browser keys that contained in the
```
jingt06: Added.

lib/web_apis/web_apis.es6.js
```diff
@@ -0,0 +1,253 @@
...
+        },
+      ],
+      code: function(browserName, browserVersion,
+        osName, osVersion, apiCatalog) {
+        let browserKey = `${browserName}_${browserVersion}_${osName}_${osVersion}`;
+        return this.browserAPIs.select(this.GroupBy.create({
+          arg1: this.BrowserAPI.BROWSER_KEY,
+          arg2: this.Count.create(),
+        })).then((browserKeyGroups) => {
+          // Exits early if given browser already exists in browserAPIs.
```
jingt06: We should only import interface/API for each browser once.

lib/web_apis/web_apis.es6.js
kgrgreer: Missing feature. Just pushed.

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
+foam.CLASS({
```
mdittmer: nit: Missing copyright header.

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
...
+    {
+      name: 'browserAPIs',
+      documentation: `The WebAPIs object where the browserAPIs will be
+        be fetched from.`,
+      typeName: 'BrowserAPI DAO',
+      required: true,
+      final: true,
+    },
+    {
+      name: 'M',
```
mdittmer: nit: Avoid starting property names with a capital letter.

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
...
+      returns: {
+        typeName: 'JSON',
+        documentation: `Returns a JSON of the form:
+            {interfaceName: {apiName: {BrowserKey: true, ...} ...} ... } ...}
+            This structure is easy for displaying a nested table.`,
+      },
+      code: function(browserKeys) {
+        return this.browserAPIs
+          .where(this.M.IN(this.BrowserAPI.BROWSER_KEY, browserKeys))
+          .select(this.M.GROUP_BY( this.BrowserAPI.INTERFACE_KEY,
```
mdittmer: nit: drop extra space before `this.BrowserAPI.INTERFACE_KEY`

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
...
+            this.ArrayDAO.create()))
```
mdittmer: This should be an `ArraySink`, no?

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
...
+                  browserAPIs[j].browserKey] = true;
+              }
+            }
+            return matrix;
+          });
+      },
+    },
+    {
+      name: 'toCSV',
+      documentation: `Takes an array of browser keys, produce a string of
```
mdittmer: nit: produces a string in CSV format

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
...
+        {
+          name: 'browserKeys',
+          typeName: 'StringArray',
+          documentation: `A string array contains valid browser keys. It will
+            be used for filtering browserAPIs.`,
+        },
+      ],
+      returns: {
+        typeName: 'JSON',
+        documentation: `Returns a JSON of the form:
```
mdittmer: nit: here and elsewhere: This is in 'returns', so it can just read "JSON of the form: [...]"

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
...
+      name: 'toMatrix',
+      documentation: `An asynchronous method that takes an array of browser
+        keys, produce a nested matrix that contains browser and interface
+        relationship. This structure should be easy to be displayed as a
+        nested table in client side.`,
+      args: [
+        {
+          name: 'browserKeys',
+          typeName: 'StringArray',
+          documentation: `A string array contains valid browser keys. It will
```
mdittmer: nit: here and other `documentation` strings: Answer *what* and *why* without repeating info the reader already has (such as "it's a string array"). E.g., `Valid browser keys used to filter browserAPIs.`.

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
...
+              }
+              table.push(row);
+            }
+          }
+          let csv = '';
+          for (let i = 0; i < table.length; i++) {
+            csv += table[i].join(',');
+            csv += '\n';
+          }
+          return(csv);
```
mdittmer: Prefer `return csv;`.

lib/web_apis/api_importer.es6.js
```diff
@@ -0,0 +1,102 @@
...
+  ],
+  properties: [
+    {
+      name: 'browserAPIs',
+      documentation: `This should be a server GCD DAO that contains
+        interface, API information for major browsers and versions
+        for recent 2 years. It is implemented as a MDAO for now.`,
+      factory: function() {
+        let browserApiDao = this.EasyDAO.create({
+          name: 'BrowserAPIDAO',
```
mdittmer: nit: `BrowserApiDAO`.

lib/web_apis/api_importer.es6.js
```diff
@@ -0,0 +1,102 @@
...
+          of: this.BrowserAPI,
+          daoType: 'MDAO',
+        });
+        browserApiDao.addPropertyIndex(this.BrowserAPI.BROWSER_KEY);
```
mdittmer: No need to change; this is fine for now.

But a controller should probably determine the make-up of this DAO eventually. That is, it will be passed in when instances of this class are constructed, and indexes that ensure acceptable performance on the analysis the controller wishes to do will be configured here.

lib/web_apis/api_importer.es6.js
```diff
@@ -0,0 +1,102 @@
...
+              osVersion,
+              interfaceName,
+              apiName,
+            }));
+          }
+        }
+      },
+    },
+  ],
+});
```
mdittmer: This component is more closing braces than code ^_^

We should keep an eye for providing a DAO to the API Extractor so that it can do this work directly whenever it hits the point in its code when it would create `apiCatalog.<interface-name>[j]`.

lib/web_apis/browser_api.es6.js
```diff
@@ -19,68 +19,82 @@
...
+      documentation: `The name of browser.`,
       required: true,
       final: true,
-      class: 'String',
     },
     {
+      class: 'String',
       name: 'browserVersion',
+      documentation: `The version of browser.`,
+      documentation: ``,
```
mdittmer: nit: Just one doc string, please ^_^

lib/web_apis/browser_api.es6.js
```diff
@@ -19,68 +19,82 @@
...
       required: true,
       final: true,
-      class: 'String',
     },
     {
+      class: 'String',
       name: 'osName',
+      documentation: `The name of operating system.`,
```
mdittmer: Here and `osVersion`: We should say something like "as reported in the  user agent string" (which is how `object-graph-js` reports it), or even "as reported by the object-graph-js library".

lib/web_apis/browser_api.es6.js
```diff
@@ -19,68 +19,82 @@
...
       name: 'osVersion',
+      documentation: `The version of operating system.`,
       required: true,
       final: true,
-      class: 'String',
     },
     {
+      class: 'String',
       name: 'interfaceName',
+      documentation: `The name of an interface. Interfaces in browser
```
mdittmer: No need to do this just yet, but we should have some kind of explainer doc (maybe even part of README.md in this repo?). The explainer should give a precise definition of "first level" and "`__proto__` object". We should have doc strings like this one link to it.

mdittmer: @jingt06 could you please file an issue to do this?

lib/web_apis/browser_api.es6.js
```diff
@@ -19,68 +19,82 @@
...
+        that have meaningful properties.`,
       required: true,
       final: true,
-      class: 'String',
     },
     {
+      class: 'String',
       name: 'apiName',
+      documentation: `The name of an API. APIs are identified as meaningful
+        properties of an interface.`,
```
mdittmer: As above: "meaningful property" should be defined in an explainer (and cited here).

lib/client/api_matrix.es6.js
jingt06: Added. I will add package license when this PR is merged to avoid any conflict.

lib/client/api_matrix.es6.js
jingt06: Changed it back to mlang.

lib/client/api_matrix.es6.js
jingt06: Done.

lib/client/api_matrix.es6.js
jingt06: Done.

lib/client/api_matrix.es6.js
jingt06: Removed.

lib/client/api_matrix.es6.js
jingt06: Changed to ArraySInk, I think both of them works though. ArrayDAO.array and ArraySink.a contains the same result..

lib/client/api_matrix.es6.js
jingt06: Done.

lib/client/api_matrix.es6.js
jingt06: Done.

lib/web_apis/api_importer.es6.js
jingt06: Done.

lib/web_apis/browser_api.es6.js
jingt06: Removed

lib/web_apis/browser_api.es6.js
jingt06: Added.

lib/web_apis/browser_api.es6.js
jingt06: Issue filed and removed the interface details.

lib/web_apis/browser_api.es6.js
jingt06: Done.

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
...
+          // Add table header.
+          table.push(['Interface', 'API'].concat(browserKeys));
+          let interfaces = Object.keys(result);
+          for (let i = 0; i < interfaces.length; i++) {
+            let interfaceName = interfaces[i];
+            let APIs = Object.keys(result[interfaceName]);
+            for (let j = 0; j < APIs.length; j++) {
+              let apiName = APIs[j];
+              let row = [interfaceName, apiName];
+              for (let browserKey in browserKeys) {
```
jingt06: Changed to for(let k = 0; ...).

lib/client/api_matrix.es6.js
```diff
@@ -0,0 +1,129 @@
...
+                if (browserKeys.hasOwnProperty(browserKey)) {
+                  if (result[interfaceName][apiName]
+                    .hasOwnProperty(browserKey)) {
+                      row.push(true);
```
jingt06: Done.

lib/web_apis/api_importer.es6.js
```diff
@@ -0,0 +1,102 @@
...
+    'org.chromium.apis.web.BrowserAPI',
+    'foam.mlang.ExpressionsSingleton',
+    'foam.dao.EasyDAO',
+  ],
+  properties: [
+    {
+      name: 'browserAPIs',
+      documentation: `This should be a server GCD DAO that contains
+        interface, API information for major browsers and versions
+        for recent 2 years. It is implemented as a MDAO for now.`,
```
jingt06: Documentation changed, removed the data availability information.

lib/client/api_matrix.es6.js
mdittmer: It's true that both will work in this case, but the interface passed to `GROUP_BY(_, sink)` is a `Sink`.

lib/client/api_matrix.es6.js
mdittmer: @arobins do you mean that visually using `Object.keys()` + `for ( let k ... )` is more consistent, or do you mean that perhaps the structure produced by `toMatrix` should produce an array of objects that have `{ browserKey: ... }` in them?

lib/web_apis/browser_api.es6.js
```diff
@@ -0,0 +1,100 @@
...
+      required: true,
+      final: true,
+    },
+    {
+      class: 'String',
+      name: 'browserKey',
+      documentation: `An unique key for this browser. Avoid the need for
+        CONCAT mLang or similar to be able to groupBy browserName,
+        browserVersion, osName, osVersion.`,
+      expression: function(browserName, browserVersion, osName, osVersion) {
```
mdittmer: The `expression` should be lazily computed once the first time its needed, and only be updated if its dependents change (which they never will). We could, instead use a `factory` that would produce the same result, but then break if we later changed one of the dependents to be non-`final`.

@kgrgreer will using `expression` trigger logic that checks for changes in dependents on every access, or is the expression recomputed via push when dependents change?

test/any/client/api_matrix-test.es6.js
```diff
@@ -0,0 +1,134 @@
...
+      'Array': ['find'],
+      'Audio': ['play'],
+    });
+    apiImporter.import('Safari', '10', 'OSX', '601', {
+      'ApplePay': ['about'],
+      'Audio': ['play', 'stop'],
+      'Array': ['find'],
+    });
+    it(`contains correct interface and API information, when all browsers
+    are selected`, function(done) {
```
mdittmer: nit: here and elsewhere: indent 2 spaces in continuation of multiline string.

test/any/web_apis/api_importer-test.es6.js
```diff
@@ -0,0 +1,72 @@
...
+    ],
+    'Function': [
+      'arguments',
+      'caller',
+    ],
+  };
+
+  apiImporter.import('Chrome', '56.0.2924.87',
+    'OSX', '10.12.2', webCatalog);
+  it('correctly imports browserAPIs to MDAO', function(done) {
```
mdittmer: nit: Drop the `M` in `MDAO`. We _should_ be able to safely assume that any DAO correctly implements the DAO interface, so which implementation we choose doesn't matter.

test/any/web_apis/api_importer-test.es6.js
```diff
@@ -0,0 +1,72 @@
...
+      {
+        interfaceName: 'Function',
+        apiName: 'arguments',
+      },
+      {
+        interfaceName: 'Function',
+        apiName: 'caller',
+      },
+    ];
+    let promises = interfaces.map((intface) => {
```
mdittmer: This is probably overkill for this test. We're not testing the soundness of the DAO query API, we're checking whether the DAO contains what we want it to contain, _and nothing more_.I was expecting something like:

```
let dao = apiImporter.browserAPIs;
let promises = [ dao.find(...), dao.find(...), ... ];
Promise.all(promises).then(function(results) {
  /* Make sure all finds are non-null */
  return dao.select(mlang.COUNT());
}).then(function(countSink) {
  /* Check that count matches promises.length */
});
```

This method has the weakness that if we mistakenly `find()` the same value twice, then our `COUNT()` check is inconsistent. However, that should be easy to spot, and in the absence of such an error, we check that the DAO doesn't contain more than it should.

test/any/web_apis/web_apis-integration.es6.js
```diff
@@ -0,0 +1,157 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+/*
+ * The object graph files under /test/data are extracted using
+ * web-apis (https://github.com/mdittmer/web-apis). The test
+ * specifications are defined at:
+ * https://docs.google.com/spreadsheets/d/1bnm-atGD2SPyQ5XBltEXOaVhzK0ey5Dvpe4Nwu2m588
```
mdittmer: This doc is not public. Don't link to it here. (Not that it contains sensitive data, but if we want to expose these specs, we should move them to a different doc).

test/any/web_apis/web_apis-integration.es6.js
```diff
@@ -0,0 +1,157 @@
...
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+/*
+ * The object graph files under /test/data are extracted using
+ * web-apis (https://github.com/mdittmer/web-apis). The test
```
mdittmer: Do you mean that they were copied from `...web-apis/data/idl`? If so, say that. The data were actually collected using `https://github.com/mdittmer/object-graph-js`.

test/any/web_apis/web_apis-integration.es6.js
```diff
@@ -0,0 +1,157 @@
...
+ * specifications are defined at:
+ * https://docs.google.com/spreadsheets/d/1bnm-atGD2SPyQ5XBltEXOaVhzK0ey5Dvpe4Nwu2m588
+ **/
+
+describe('WebAPI and api extractor', function() {
+  let chrome56 = global.DATA.chrome56;
+  let edge14 = global.DATA.edge14;
+  let safari602 = global.DATA.safari602;
+  let og = global.ObjectGraph;
+  let extractor = org.chromium.apis.web.apiExtractor.create({});
```
mdittmer: Does this import take more than 10s? If not, we should declare these variables here, but let's set them and import the data in a `beforeEach()`. That way, if at some point we write a test that mucks around with the internal state of these variables it won't affect other tests. If so, _maybe_ we should leave it as-is: on one hand, we don't want tests to take too long to run; on the other hand, this is an _integration test suite_ so it's okay for it to be a little sluggish.

@arobins WDYT?

mdittmer: If we do leave it as-is, we should add a comment somewhere conspicuous that all tests in this suite must be `DATA` read-only.

test/any/web_apis/web_apis-integration.es6.js
```diff
@@ -0,0 +1,157 @@
...
+      expect(webCatalogMap.BiquadFilterNode.Q).toEqual({
+        Chrome_56_Windows_10: true,
+        Edge_14_Windows_10: true,
+        Safari_602_OSX_10: true,
+      });
+      done();
+    });
+    describe('Window interface', function() {
+      it(`contains first level objects`, function(done) {
+        // Constant value are not identified as all-capital value.
```
mdittmer: Copy/paste error? Shouldn't this say something about first-level objects that are referenced in multiple places? Describe the edge case we're trying to catch here.

test/any/web_apis/web_apis-integration.es6.js
```diff
@@ -0,0 +1,157 @@
...
+        });
+        expect(webCatalogMap.Window.MouseEvent).toEqual({
+          Chrome_56_Windows_10: true,
+          Edge_14_Windows_10: true,
+          Safari_602_OSX_10: true,
+        });
+        done();
+      });
+      it('does not contain non-interface object', function(done) {
+        // Chrome does not expose FontFase as global interface.
```
mdittmer: nit: `FontFace`.

test/any/web_apis/web_apis-integration.es6.js
```diff
@@ -0,0 +1,157 @@
...
+          Safari_602_OSX_10: true,
+        });
+        expect(webCatalogMap.Window.MouseEvent).toEqual({
+          Chrome_56_Windows_10: true,
+          Edge_14_Windows_10: true,
+          Safari_602_OSX_10: true,
+        });
+        done();
+      });
+      it('does not contain non-interface object', function(done) {
```
mdittmer: nit: `objects`.

test/any/web_apis/web_apis-integration.es6.js
mdittmer: I don't understand the rationale for this test. What's the edge case we're testing for? Testing that something that doesn't exist indeed doesn't exist seems vacuous.

test/any/web_apis/web_apis-integration.es6.js
```diff
@@ -0,0 +1,157 @@
...
+        // Chrome does not expose FontFase as global interface.
+        expect(webCatalogMap.Window.FontFaceSet.
+          Chrome_56_Windows_10).toBeUndefined;
+        done();
+      });
+    });
+    it('filters out built-in function object properties', function(done) {
```
mdittmer: Suggested rename: `ignores built-in function properties on function instances`

test/any/web_apis/web_apis-integration.es6.js
arobins: Loading it every time would definitely be better. Test run time is also important, but I'd rather have slightly longer hermetic tests than faster and potentially flaky (or worse, incorrect) tests.

test/any/client/api_matrix-test.es6.js
```diff
@@ -0,0 +1,134 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+describe('ApiMatrix', function() {
+  describe('toMatrix()', function() {
```
jingt06: Add test that pass an empty array as browserKeys.

test/any/client/api_matrix-test.es6.js
```diff
@@ -0,0 +1,134 @@
...
+    let apiImporter = org.chromium.apis.web.ApiImporter.create();
+    let apiMatrix = org.chromium.apis.web.ApiMatrix.create({
+      browserAPIs: apiImporter.browserAPIs,
+    });
+    apiImporter.import('Chrome', '55', 'Window', '10', {
```
jingt06: Dropped, using dao.put(...) instead.

test/any/client/api_matrix-test.es6.js
jingt06: Done.

test/any/client/api_matrix-test.es6.js
```diff
@@ -0,0 +1,134 @@
...
+        });
+        expect(interfaceMap.Audio.play).toEqual({
+          Edge_14_Window_10: true,
+          Safari_10_OSX_601: true,
+        });
+        expect(interfaceMap.Audio.stop).toEqual({
+          Chrome_55_Window_10: true,
+          Safari_10_OSX_601: true,
+        });
+        done();
```
jingt06: done dropped. I thought this is asynchronous since it is inside a then.

test/any/client/api_matrix-test.es6.js
```diff
@@ -0,0 +1,134 @@
...
+        expect(interfaceMap.Array.find).toEqual({
+          Chrome_55_Window_10: true,
+          Edge_14_Window_10: true,
+        });
+        expect(interfaceMap.Audio.play).toEqual({
+          Edge_14_Window_10: true,
+        });
+        expect(interfaceMap.Audio.stop).toEqual({
+          Chrome_55_Window_10: true,
+        });
```
jingt06: Added

test/any/client/api_matrix-test.es6.js
```diff
@@ -0,0 +1,134 @@
...
+        done();
+      });
+    });
+  });
+
+  describe('toCSV()', function() {
+    let apiImporter = org.chromium.apis.web.ApiImporter.create();
```
jingt06: Added beforeEach function to create a new ApiMatrix each time.

test/any/client/api_matrix-test.es6.js
```diff
@@ -0,0 +1,134 @@
...
+      'Array': ['find'],
+    });
+    it(`produces correct csv string when all browsers are selected`,
+    function(done) {
+      apiMatrix.toCSV([
+        'Chrome_55_Window_10',
+        'Edge_14_Window_10',
+        'Safari_10_OSX_601',
+      ]).then((csvStr) => {
+        expect(csvStr.split('\n').sort()).toEqual(
```
jingt06: Add sort in every test. Sort is added so that the ordering of row does not affect to the result.

test/any/client/api_matrix-test.es6.js
```diff
@@ -0,0 +1,134 @@
...
+        'Chrome_55_Window_10',
+        'Edge_14_Window_10',
+      ]).then((csvStr) => {
+        expect(csvStr).toEqual(
+          'Interface,API,Chrome_55_Window_10,Edge_14_Window_10\n' +
+          'Array,find,true,true\n' + 'Audio,stop,true,false\n' +
+          'Audio,play,false,true\n'
+        );
+        done();
+      });
```
jingt06: Add test for header line.
Added test for unknown keys. The unknown key is still listed in the header, with all false value.

test/any/web_apis/api_importer-test.es6.js
jingt06: Done.

test/any/web_apis/api_importer-test.es6.js
jingt06: Done.

test/any/web_apis/web_apis-integration.es6.js
jingt06: Done

test/any/web_apis/web_apis-integration.es6.js
jingt06: Yes, from data/og.
Change documentation to 'collected using object-graph-js'.

test/any/web_apis/web_apis-integration.es6.js
jingt06: Added beforeEach function to create interface matrix for each tests.

test/any/web_apis/web_apis-integration.es6.js
jingt06: Copy paste error, fixed now.
Test for multiple first-level objects referenced to the same object added. (MediaStream vs webkitMediaStream in Chrome).

test/any/web_apis/web_apis-integration.es6.js
jingt06: Done

test/any/web_apis/web_apis-integration.es6.js
jingt06: Done

jingt06: Just to make sure non-interface namespace will not be included as a exposed interface under window.

test/any/web_apis/web_apis-integration.es6.js
jingt06: Done

test/any/client/api_matrix-test.es6.js
mdittmer: That messes with the header row, no?

test/any/client/api_matrix-test.es6.js
```diff
@@ -17,91 +17,161 @@
...
     });
-    apiImporter.import('Edge', '14', 'Window', '10', {
-      'Array': ['find'],
-      'Audio': ['play'],
-    });
-    apiImporter.import('Safari', '10', 'OSX', '601', {
-      'ApplePay': ['about'],
-      'Audio': ['play', 'stop'],
-      'Array': ['find'],
+    browserApiDao.put(org.chromium.apis.web.BrowserAPI.create({
```
mdittmer: The "height" of this setup code almost makes me want to undo my request for it ^_^

We should keep the dao create-then-put setup flow, but let's add a helper method outside the `beforeEach()` that constructs these objects. Then all these lines become something like `browserApiDao.put(browserApi('Chrome', '55', 'Windows', '10', 'Array', 'find'))`.

test/any/client/api_matrix-test.es6.js
```diff
@@ -17,91 +17,161 @@
...
+      browserName: 'Chrome',
+      browserVersion: '55',
+      osName: 'Window',
```
mdittmer: nit: here and elsewhere. The OS is called `Windows`.

test/any/client/api_matrix-test.es6.js
mdittmer: Let's use `Object.keys(...).sort()` to get the groups in a deterministic order. We can also use `dao.orderBy(...)` on the query before `select()` to get ordering within groups.

jingt06:
Add ORDER_BY(INTERFACE_KEY) to ensure the returned CSV are in alphabet order. 
Removed .split(...).sort() in api_matrix test.

lib/web_apis/browser_api.es6.js
kgrgreer: Expressions are invalidated via push when dependent values change, then recalculated when accessed next.

PR #21, opened 2017-03-15
Node and webpack setup

jingt06:
Used the same pattern as web-apis.
`npm run serve` start the server.

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -1,0 +1,53 @@
+'use strict';
+
+const HtmlWebpack = require('html-webpack-plugin');
+const path = require('path');
+const webpack = require('webpack');
+const ChunkWebpack = webpack.optimize.CommonsChunkPlugin;
+
+const rootDir = path.resolve(__dirname, '..');
+
+const FOAM_DIR = `${__dirname}/../node_modules/foam2`;
+const execSync = require('child_process').execSync;
+execSync(`node ${FOAM_DIR}/tools/build.js web`);
+
+module.exports = {
+  entry: {
+    foam: [path.resolve(rootDir, 'node_modules/foam2', 'foam-bin')],
+  },
+  output: {
+    filename: '[name].bundle.js',
+    path: path.resolve(rootDir, 'static/bundle'),
+  },
+  module: {
+    loaders: [
+      {
+        test: /\.es6\.js$/,
+        loader: 'babel-loader',
+        query: {
+          presets: ['es2015'],
+          plugins: ['transform-runtime'],
+        },
+      },
+    ],
+  },
+  plugins: [
+    new ChunkWebpack({
+      filename: 'vendors.bundle.js',
+      minChunks: Infinity,
+      name: 'vendors',
+    }),
+    new HtmlWebpack({
+      filename: 'index.html',
+      inject: 'body',
+      template: path.resolve(rootDir, 'static', 'index.html'),
+    })
+  ],
+  resolve: {
+    extensions: ['.js'],
+  },
+  node: {
+    fs: 'empty',
+    dns: 'empty',
+  },
+};
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -1,4 +1,20 @@
-'use strict';
+/**
+ * @license
+ * Copyright 2017 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


config/webpack.config.js
```diff
@@ -0,0 +1,69 @@
+/**
```
mdittmer: Let's discuss the need for this. If the entry point is `foam-bin` and the only other thing we do is fix script tags in an html file, I think I'd prefer to not use `webpack`. It's bundling code that's already bundled by FOAM. We can just as easily copy `foam-bin` to `static/bundle` and refer to the correct location from `html` file(s).

main/serve.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+let express = require('express');
```
mdittmer: I don't think we're doing anything complex enough to warrant all these third party dependencies.

For serving let's use a [`Server`](https://github.com/foam-framework/foam2/blob/master/src/foam/net/node/Server.js) and [`StaticFileHandler`](https://github.com/foam-framework/foam2/blob/master/src/foam/net/node/StaticFileHandler.js) (and later add a [`RestDAOHandler`](https://github.com/foam-framework/foam2/blob/master/src/foam/net/node/RestDAOHandler.js)).

Also, we can parse a port number ourselves (and give reasonable error + help output). If we end up needing a bunch of params, we can revisit using `yargs` or simply model the params in FOAM.

package.json
```diff
@@ -18,8 +19,10 @@
   "license": "Apache-2.0",
   "homepage": "https://github.com/GoogleChrome/confluence#readme",
   "dependencies": {
+    "express": "^4.15.2",
```
mdittmer: See above discussion: change in deps probably unnecessary.

scripts/serve.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/zsh
```
mdittmer: We may drop this if we drop the dependency on `webpack`.

static/index.html
```diff
@@ -0,0 +1,8 @@
+<!DOCTYPE html>
```
mdittmer: Let's refer directly to our assets. If we need `webpack` at all, it will be to refer to a known bundle name for our own code.

main/serve.js
```diff
@@ -16,25 +16,14 @@
...
+  port: 9000,
+});
 
-app.use(express.static('static/bundle'));
+server.exportFile('/', `${__dirname}/../static/index.html`);
 
-app.get('/', function(req, res) {
-  res.sendFile(path.join(__dirname, '../', 'static', 'index.html'));
-});
+server.exportDirectory('/', `${__dirname}/../static/bundle`);
```
mdittmer: Does exporting both a file and a directory over the same path work?

config/webpack.config.js
```diff
@@ -0,0 +1,63 @@
...
+const webpack = require('webpack');
+const ChunkWebpack = webpack.optimize.CommonsChunkPlugin;
+
+const rootDir = path.resolve(__dirname, '..');
+
+const FOAM_DIR = `${__dirname}/../node_modules/foam2`;
+const execSync = require('child_process').execSync;
+execSync(`node ${FOAM_DIR}/tools/build.js  web,gcloud`);
+
+module.exports = {
```
mdittmer: Let's not use webpack for this. We can `execSync('mv ${FOAM_DIR}/foam-bin.js ${rootDir}/static/bundle/foam.bundle.js');`.

Also, let's make our string constants consistent (and maybe also create a "bundle dir" constant for copying FOAM?). I prefer consistently `NAMED_LIKE_THIS` for string constants, but `namedLikeThis` is okay too (since `const` tells us they are const).

This should eliminate the need for webpack-related stuffs as devDependencies, right?

main/serve.js
jingt06: Yes, it works.
> Handlers handle requests in the order they appear in the array.

When '/' is handled by the fileHandler, it is not passed to the staticFileHandler. There is no conflict.

jingt06:
Webpack is not used for packaging multiple files for now. But as I ship more client side modules, it will be helpful.
Without webpack, I can only export every directory that have client side files and include them in html files, which will be hard to maintain.

mdittmer:
What's going on with https://travis-ci.org/GoogleChrome/confluence/builds/212168449 ?

mdittmer:
Does this need a `merge master` and `push` to go green?

PR #51, opened 2017-04-19
Fix code coverage metrics and test initialization code

mdittmer:
This change includes:

- Run tests (and coverage) in browser and Node (and combine results)
- Fix test initializations that were running too early
- Separate module (code) dependency injection from test data injection
- Add missing angular dependency to angular code
- Introduce shared browser/Node code coverage thresholds (currently lower
  than desired; that's a TODO)


config/jasmine.json
```diff
@@ -8,6 +8,7 @@
   ],
   "helpers": [
     "node/**/*-helper*.js",
-    "any/**/*-helper*.js"
+    "any/**/*-helper*.js",
+    "../node_modules/foam2/test/helpers/testcontext.js"
```
arobins: I recommend using trailing commas, if possible. It makes the diff for appending a new entry to the list easier to review.

package.json
```diff
@@ -3,9 +3,13 @@
   "version": "0.0.1",
   "description": "Web API Confluence Dashboard",
   "scripts": {
-    "coverage": "karma start ./config/karma.coverage.conf.js",
-    "test": "karma start ./config/karma.all.conf.js",
-    "serve": "zsh scripts/serve.sh"
+    "coverage": "npm run coverageNode && npm run coverageWeb && istanbul-combine -d .coverage -p summary -r lcov -r html -r json -b . .node_coverage/coverage.json .web_coverage/*/coverage*.json && istanbul check-coverage --config config/istanbul.yml \".coverage/**/coverage*.json\"",
```
arobins: Is there any way to add some linebreaks to this? I think it would be easier to read with breaks before or after the &&s.

test/any/data-helper.js
```diff
@@ -0,0 +1,18 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+'use strict';
+
+// Load external libraries and data.
+let libs = {
+  ObjectGraph: require('object-graph-js').ObjectGraph,
+  DATA: {
+	chrome56: require('../data/window_Chrome_56_Windows_10.0.json'),
```
arobins: Spaces for indentation

test/any/files-helper.js
```diff
@@ -0,0 +1,64 @@
...
+// Manually require code under test (in alphabetical order).
+//
+// TODO(markdittmer): Move to FOAM classloader for this.
+
+beforeAll(function() {
+  // Use of foam.isServer = <running in Node JS?>:
+  // - Do not load Angular-related in Node JS (foam.isServer = true).
+  // - Do not load import-from-filesystem in browser.
+
+  if (!foam.isServer) {
```
arobins: Maybe having two subhelpers for importing server-only and client-only stuff would be a better structure?

test/any/files-helper.js
```diff
@@ -0,0 +1,64 @@
...
+    // TODO(markittmer): Fix: Alphabetical order exception: main/app defines
+    // angular module used by other code.
```
arobins: Is the angular module only required by stuff imported directly after this? If it is, document that the dependent code is entirely in this block. If not, I think it would be better to break the overall alphabetical ordering and move those imports here, and then document it.

test/any/web_catalog/api_extractor-integration.es6.js
```diff
@@ -18,325 +18,330 @@
...
-    'key': 'window',
-  });
-  let extractor = org.chromium.apis.web.ApiExtractor.create({});
+  let objectGraph;
+  let og;
+  let extractor;
+  let apiCatalog;
+  beforeEach(function() {
+    objectGraph = global.ObjectGraph;
+    og = objectGraph.fromJSON({
```
arobins: Just formatting?

config/jasmine.json
mdittmer: I believe this file needs to be valid JSON. I would, otherwise, agree.

package.json
mdittmer: Again, I think we're stuck with this because of the requirement for valid JSON.  We could punt to a shell script. IMHO that adds unnecessary indirection. @arobins WDYT?

test/any/data-helper.js
mdittmer: Done (impending push).

test/any/web_catalog/api_extractor-integration.es6.js
mdittmer: Kinda-sorta. Initialization needed to be moved into a `beforeEach()` call. That said, the initialization data/logic has not changed.

config/jasmine.json
arobins: But the configuration files below don't need to be valid JSON?

package.json
arobins: I don't think moving it to a script helps.

test/any/web_catalog/api_extractor-integration.es6.js
arobins: Right. I meant specifically about the content of this statement.

config/jasmine.json
mdittmer: That's right. I believe Jasmine will `JSON.parse()` this file, whereas Karma loads configuration files by `require()`ing them as Node JS modules.

PR #52, opened 2017-04-25
Test aggressive removal

mdittmer:
In addition to test code, this PR contains a fix for the aggressive removal algorithm, and models sinks passed to `someDAO.select()` to make sure the underlying implementation does choke on a missing sink method.

### DIFF
--- a/lib/client/api_matrix.es6.js
+++ b/lib/client/api_matrix.es6.js
@@ -11,12 +11,13 @@ foam.CLASS({
     to CSV format string or JSON matrix that can be displayed on HTML as
     a nested table.`,
   requires: [
-    'org.chromium.apis.web.Browser',
-    'org.chromium.apis.web.WebInterface',
-    'org.chromium.apis.web.BrowserWebInterfaceJunction',
-    'foam.dao.EasyDAO',
+    'foam.dao.AnonymousSink',
     'foam.dao.ArraySink',
+    'foam.dao.EasyDAO',
     'foam.mlang.ExpressionsSingleton',
+    'org.chromium.apis.web.Browser',
+    'org.chromium.apis.web.BrowserWebInterfaceJunction',
+    'org.chromium.apis.web.WebInterface',
   ],
   properties: [
     {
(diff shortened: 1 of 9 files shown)


lib/confluence/aggressive_removal.es6.js
```diff
@@ -66,58 +67,70 @@ foam.CLASS({
...
-              promises.push(prevReleaseBrowsers[i].interfaces.select({
-                put: function(_, iface) {
-                  if (!removedIfaces.hasOwnProperty(iface.interfaceKey)) {
-                    removedIfaces[iface.interfaceKey] = 0;
-                  }
-                },
-              }));
+              promises.push(prevReleaseBrowsers[i].interfaces.select(
+                  this.AnonymousSink.create({sink: {put: function(_, iface) {
+                    if (!removedIfaces.hasOwnProperty(iface.interfaceKey)) {
```
arobins: You don't need the check. Setting it to 0 a whole bunch of times should be safe.

lib/confluence/metric_computer.es6.js
```diff
@@ -102,25 +102,27 @@ foam.CLASS({
...
+      name: 'run',
+      documentation: `Computes the Metric Result for each major browsers
+          contained in the given browserApiDAO.`,
       code: function() {
-        this.browserDAO.select(
+        return this.browserDAO.select(
             this.mlang.GROUP_BY(this.Browser.BROWSER_NAME, this.mlang.GROUP_BY(
                 this.Browser.OS_NAME,
                 this.mlang.COUNT())))
         .then((groups) => {
```
arobins: nit: this call should be indented +4

lib/client/api_matrix.es6.js
```diff
@@ -11,12 +11,13 @@ foam.CLASS({
     to CSV format string or JSON matrix that can be displayed on HTML as
     a nested table.`,
   requires: [
-    'org.chromium.apis.web.Browser',
-    'org.chromium.apis.web.WebInterface',
-    'org.chromium.apis.web.BrowserWebInterfaceJunction',
-    'foam.dao.EasyDAO',
+    'foam.dao.AnonymousSink',
```
mdittmer: [It's here](https://github.com/foam-framework/foam2/blob/073884a6e4e3900c48f4aebaf664447235dfd431/src/foam/dao/Sink.js#L181). It allows for providing a partial `Sink` implementation and fills in noops for other methods.

arobins: Ah, I think I ended up in the foam repo instead of the foam2 repo by following links from the first search result for FOAM framework.

mdittmer: Yeah... We need more "You should be looking here..." stuff in our READMEs and other places.

PR #53, opened 2017-05-03
Sink for set operation: SET_MINUS(subtrahendDAO)

mdittmer:
This abstraction should greatly simplify the business logic in `lib/confluence` metrics.

### DIFF
--- a/lib/confluence/set_ops.es6.js
+++ b/lib/confluence/set_ops.es6.js
@@ -8,29 +8,41 @@ foam.CLASS({
   name: 'SetMinus',
   extends: 'foam.dao.ProxySink',
 
+  documentation: `Put incoming objects to delegate iff the same object (by id)
+      is not found in "subtrahend" DAO.`,
+
   properties: [
     {
       class: 'foam.dao.DAOProperty',
+      documentation: `"subtrahend" in set operation:
+          SET_MINUS(minuend, subtrahend)`,
       name: 'subtrahend',
       required: true,
     },
     {
       name: 'next_',
+      documentation: `Next Promise to wait for before committing an incoming
+          put().`,
       factory: () => Promise.resolve(),
     },
     {
+      documentation: `Promise to be resolved when all put()s have been dropped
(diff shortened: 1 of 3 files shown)


lib/confluence/set_ops.es6.js
```diff
@@ -0,0 +1,92 @@
...
+
+foam.CLASS({
+  package: 'org.chromium.mlang.sink',
+  name: 'SetMinus',
+  extends: 'foam.dao.ProxySink',
+
+  properties: [
+    {
+      class: 'foam.dao.DAOProperty',
+      name: 'subtrahend',
```
arobins: I learned a new word today

test/any/confluence/set_ops-test.es6.js
```diff
@@ -0,0 +1,248 @@
...
+            expect(array.length).toBe(10);
+            for (var i = 0; i < 10; i++) {
+              expect(array[i].id).toBe(i);
+            }
+
+            done();
+          });
+    });
+
+    it('should yield full set on SET_MINUS(nullSet)', done => {
```
arobins: Duplicated test

test/any/confluence/set_ops-test.es6.js
```diff
@@ -0,0 +1,248 @@
...
+        if ((i % 2) === 0) {
+          subtrahend.put(mkNum(i));
+        }
+      }
+
+      set.orderBy(Num.ID).select(E.SET_MINUS(subtrahend))
+          .then(setMinusSink =>setMinusSink.getSink())
+          .then(sink => {
+            var array = sink.a;
+            expect(array.length).toBe(5);
```
arobins: I assume in JS array equality isn't a thing? Because if it is, array === [1,3,5,7,9] feels like a much easier to follow check. Possibly same for the [0..10] case above, though that array is longer.

mdittmer: It is; pushed simplified expectations. PTAL.

PR #54, opened 2017-05-04
Test and implement set intersection

mdittmer:
Abstracts async binary set ops into common base class. Also introduces one new SET_MINUS test for composition and a bug fix: forward eof() to delegate.

### DIFF
--- a/lib/confluence/set_ops.es6.js
+++ b/lib/confluence/set_ops.es6.js
@@ -5,18 +5,17 @@
 
 foam.CLASS({
   package: 'org.chromium.mlang.sink',
-  name: 'SetMinus',
+  name: 'AsyncSink',
   extends: 'foam.dao.ProxySink',
 
-  documentation: `Put incoming objects to delegate iff the same object (by id)
-      is not found in "subtrahend" DAO.`,
+  documentation: 'Common pattern for sinks that wait on async operations.',
 
   properties: [
     {
       class: 'foam.dao.DAOProperty',
-      documentation: `"subtrahend" in set operation:
-          SET_MINUS(minuend, subtrahend)`,
-      name: 'subtrahend',
+      documentation: `"secondary" in binary set operation:
+          OPERATION(primary, secondary)`,
+      name: 'secondary',
       required: true,
     },
     {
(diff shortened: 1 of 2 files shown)


test/any/confluence/set_ops-test.es6.js
```diff
@@ -206,4 +231,158 @@ describe('Set ops', () => {
...
+      set.orderBy(Num.ID).select(E.INTERSECT(secondary))
+          .then(sink => {
+            expect(sink.a.map(num => num.id))
+                .toEqual([0, 2, 4, 6, 8]);
+
+            done();
+          });
+    });
+
+    it('should remove some items when primary is subset', done => {
```
arobins: Exclude might be a better word than remove, since if you are thinking about it in terms of the mutation applied to primary, the intersect operation is not doing anything.

test/any/confluence/set_ops-test.es6.js
```diff
@@ -206,4 +231,158 @@ describe('Set ops', () => {
...
+      set.orderBy(Num.ID).select(E.INTERSECT(secondary))
+          .then(sink => {
+            expect(sink.a.map(num => num.id))
+                .toEqual([0, 2, 4, 6, 8]);
+
+            done();
+          });
+    });
+
+    it('should reject when primary rejects select()', done => {
```
arobins: Since this is a property of the parent class, it's probably not providing much value to test it on each subclass.

lib/confluence/set_ops.es6.js
```diff
@@ -94,19 +95,53 @@ foam.CLASS({
...
+foam.CLASS({
+  package: 'org.chromium.mlang.sink',
+  name: 'Intersect',
+  extends: 'org.chromium.mlang.sink.AsyncSink',
+
+  documentation: `Put incoming objects to delegate iff the same object (by id)
+      is found in "secondary" DAO.`,
+
+  listeners: [
+    function onSecondaryFind(sub, o, found) {
```
arobins: As discussed offline, I found the logic flow here difficult to follow. I think it could use some documentation.

mdittmer:
Refactored inheritance/composition of `AsyncSink` and its helpers. Also addressed outstanding comments. @arobins PTAL.

lib/confluence/set_ops.es6.js
```diff
@@ -94,19 +122,68 @@ foam.CLASS({
 });
 
 foam.CLASS({
+  package: 'org.chromium.mlang.sink',
+  name: 'SetMinus',
```
arobins: Maybe SetMinusDecider?

test/any/confluence/set_ops-test.es6.js
```diff
@@ -178,32 +232,132 @@ describe('Set ops', () => {
...
+    });
+
+    it('should exclude some items when primary is subset', done => {
+      var set = MDAO.create({of: Num});
+      var secondary = MDAO.create({of: Num});
+
+      for (var i = 0; i < 10; i++) {
+        if ((i % 2) === 0) {
+          set.put(mkNum(i));
+        secondary.put(mkNum(i));
```
arobins: Move this down a line.

PR #55, opened 2017-05-04
Implement UNION set operation as UnionDAO

mdittmer:
Chose to do this one as a DAO instead of a sink because the secondary introduces new data that are not backed by the primary. I.e., it _feels wrong_ to implement something that supports:

```js
foo.select(UNION(bar))
```

where the data `put()` to the sink would include items that are not contained in `foo`.

### DIFF
--- a/lib/confluence/set_ops.es6.js
+++ b/lib/confluence/set_ops.es6.js
@@ -187,3 +187,113 @@ foam.CLASS({
     },
   ],
 });
+
+foam.CLASS({
+  package: 'org.chromium.dao',
+  name: 'UnionDAO',
+  extends: 'foam.dao.AbstractDAO',
+  implements: ['foam.mlang.Expressions'],
+
+  documentation: 'Read-only DAO that captures the union of two delegate DAOs.',
+
+  classes: [
+    {
+      name: 'NoEOFSink',
+      extends: 'foam.dao.ProxySink',
+
+      methods: [function eof() {}],
+    },
+  ],
+
+  properties: [
+    {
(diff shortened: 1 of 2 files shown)


lib/confluence/set_ops.es6.js
```diff
@@ -158,13 +158,136 @@ foam.CLASS({
...
+
+        // Select primary -> "sink"-with-no-eof() callback.
+        this.primary.select(this.NoEOFSink.create({delegate: sink}))
+            .then(onSelect, onError);
+        // Select secondary \ primary -> "sink"-with-no-eof() callback.
+        this.secondary.select(
+            this.SET_MINUS(this.primary, this.NoEOFSink.create({
+              delegate: sink,
+            })))
+                .then(onSelect, onError);
```
arobins: This indentation feels weird. I think you want to either tack it to the end of the select, or indent +4 from the start of the statement.

lib/confluence/set_ops.es6.js
```diff
@@ -158,13 +158,136 @@ foam.CLASS({
...
+      });
+    },
+    function listen(sink, skip, limit, order, predicate) {
+      if (skip || limit || order || predicate)
+        throw new Error('UnionDAO does not support skip/limit/order/predicate');
+
+      // Listen to primary.
+      this.primary.listen(sink);
+      // Listen to secondary \ primary.
+      this.primary.listen(this.SET_MINUS(this.primary, sink));
```
arobins: secondary

PR #56, opened 2017-05-05
Set ops changes: Helper methods nary instead of binary + two bug fixes

mdittmer:
Bug fixes:

- Better `.then()`ing to defeat race condition in `AsyncSync`s
- Fix logic error in resolving `UnionDAO.find()`

### DIFF
--- a/lib/confluence/set_ops.es6.js
+++ b/lib/confluence/set_ops.es6.js
@@ -81,15 +81,19 @@ foam.CLASS({
     },
     function put(sub, o) {
       if (this.rejected_) return;
-      this.next_ = this.next_.then(
-          this.onNext.bind(this, sub, o),
-          this.onError);
+      this.next_ = this.next_
+          .then(() => this.secondary.find(o))
+          .then(found => this.decider.shouldPutToDelegate(o, found) ?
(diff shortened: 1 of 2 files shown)


mdittmer:
Note: Depends on https://github.com/foam-framework/foam2/pull/317 for passing tests.

lib/confluence/set_ops.es6.js
```diff
@@ -295,24 +293,46 @@ foam.CLASS({
 
   methods: [
     function SET_MINUS(dao, sink) {
+      if (Array.isArray(dao))
```
arobins: It might be easier to follow/maintain if you handled the single case by wrapping in array. That way, the sink initialization stays in one place.

lib/confluence/set_ops.es6.js
```diff
@@ -295,24 +293,46 @@ foam.CLASS({
...
+        return this.cascadeSink_(this.IntersectDecider, dao, sink);
```
arobins: SetMinusDecider, not IntersectDecider

arobins: Also this PR should have some tests added because the existing tests weren't sufficient to catch this problem.

PR #57, opened 2017-05-05
Implement aggressive_removal metric in terms of set operations

mdittmer:
Depends on https://github.com/foam-framework/foam2/pull/317 #56 for passing tests.

### DIFF
--- a/lib/confluence/aggressive_removal.es6.js
+++ b/lib/confluence/aggressive_removal.es6.js
@@ -47,14 +47,13 @@ foam.CLASS({
       name: 'computeForBrowser',
       documentation: 'Compute aggressive removal value for a single browser.',
       code: function(browsers, date, browser) {
-        let browserKeys = browsers.map((browser) => browser.browserKey);
+        let otherBrowserIDs = browsers.map(brwsr => brwsr.browserKey)
+            .filter(id => !foam.util.equals(id, browser.id));
         return Promise.all([
           // Current browsers from vendor other than "browser".
-          this.browserDAO.where(this.mlang.AND(
-              this.mlang.NEQ(this.Browser.BROWSER_NAME, browser.browserName),
-              this.mlang.IN(this.Browser.BROWSER_KEY, browserKeys)))
-                  .orderBy(this.mlang.DESC(this.Browser.RELEASE_DATE))
-                  .select(),
+          this.browserDAO.where(this.mlang.IN(this.Browser.ID, otherBrowserIDs))
+              .orderBy(this.mlang.DESC(this.Browser.RELEASE_DATE))
+              .select(),
           // Browsers released before the grace period "browser" vendor.
           this.browserDAO.where(this.mlang.AND(
                   this.mlang.EQ(this.Browser.BROWSER_NAME, browser.browserName),
--- a/lib/confluence/aggressive_removal.es6.js
+++ b/lib/confluence/aggressive_removal.es6.js
@@ -118,7 +118,7 @@ foam.CLASS({
                         this.mlang,
                         prevLessOne.map(brwsr => brwsr.interfaces)),
                     this.mlang.SET_MINUS(
-                        postRemoval.interfaces, browser.interfaces,
+                        [postRemoval.interfaces, browser.interfaces],
                         // Count APIs in set.
                         this.mlang.COUNT()))))
             // Store count.
--- a/lib/confluence/aggressive_removal.es6.js
+++ b/lib/confluence/aggressive_removal.es6.js
@@ -103,25 +103,26 @@ foam.CLASS({
         let firstCurrent = currentLessOne.shift();
 
         // INTERSECT() decorates sink from a select(), so to compute
-        // intersection-of-current-releases, use
-        // first-current-release.select(INTERSECT(rest-of-current-releases)).
+        // intersection-of-<foos>, use first-foo.select(INTERSECT(rest-foos)).
         return firstCurrent.interfaces.select(
-            // Intersect with interfaces of other browers in "current" to get
-            // interfaces that appear in all "current" browsers.
+            // (1) APIs in "current" from other vendors and old releases of
+            //     "browser":
+            //     INTERSECT(<interfaces for each item in "current">,
+            //         UNION(<interfaces of "browser" relases before grace
+            //             period>))
             this.mlang.INTERSECT(
-                currentLessOne.map(brwsr => brwsr.interfaces),
-                // Intersect with removed-before-grace-period, not-added-back:
-                // {...} INTERSECT (allAPIsOlderThanPostRemoval \
-                //                  postRemovalAPIs \ browserAPIs).
-                this.mlang.INTERSECT(
-                    this.mlang.UNION.apply(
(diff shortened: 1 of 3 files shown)


lib/confluence/aggressive_removal.es6.js
```diff
@@ -43,97 +39,110 @@ foam.CLASS({
...
-                      numAggressiveRemoval,
-                      date,
-                    }));
-              });
-            });
-          }));
-        }
-        return Promise.all(computeForBrowsers);
+        // Select from first current browser.
+        return firstCurrent.interfaces.select(
```
arobins: I think the reason that this is handled specially should be documented.

lib/confluence/aggressive_removal.es6.js
```diff
@@ -43,97 +39,110 @@ foam.CLASS({
...
+            // Intersect with other current browsers to get APIs shipped by
+            // all (non-"browser" vendor) current browsers.
```
arobins: Suggestion: Intersect with interfaces of other browsers in "current" to get a list of interfaces that exist in all of them.

lib/confluence/aggressive_removal.es6.js
```diff
@@ -43,97 +39,110 @@ foam.CLASS({
...
+    },
+    {
+      name: 'computeFromBrowsers',
+      documentation: `Compute aggressive removal value for "browser", given
+          arrays of "current" browsers from other vendors and "previous"
+          browsers (from "browser" vendor) that are older than the grace
+          period.
+
+          Let browserAPIs = <"browser" interfaces>
+          Let curAPIs = INTERSECT(<"current" interfaces>)
```
arobins: suggestion: interfaces for each item in "current"

lib/confluence/aggressive_removal.es6.js
```diff
@@ -43,97 +39,110 @@ foam.CLASS({
...
+      name: 'computeForBrowser',
+      documentation: 'Compute aggressive removal value for a single browser.',
+      code: function(browsers, date, browser) {
+        let browserKeys = browsers.map((browser) => browser.browserKey);
+        return Promise.all([
+          // Current browsers from vendor other than "browser".
+          this.browserDAO.where(this.mlang.AND(
+              this.mlang.NEQ(this.Browser.BROWSER_NAME, browser.browserName),
+              this.mlang.IN(this.Browser.BROWSER_KEY, browserKeys)))
+                  .orderBy(this.mlang.DESC(this.Browser.RELEASE_DATE))
```
arobins: How does this produce just the latest version for each browser key? Is it missing a group by somewhere?

browserDAO has a function getLatestBrowserFromEachVendorAtDate. Is it possible to use that here?

lib/confluence/aggressive_removal.es6.js
```diff
@@ -43,97 +39,110 @@ foam.CLASS({
...
+            browser => this.computeForBrowser(browsers, date, browser)));
+      },
+    },
+    {
+      name: 'computeForBrowser',
+      documentation: 'Compute aggressive removal value for a single browser.',
+      code: function(browsers, date, browser) {
+        let browserKeys = browsers.map((browser) => browser.browserKey);
+        return Promise.all([
+          // Current browsers from vendor other than "browser".
```
arobins: As discussed offline, I think we need a terminology refresh here. 'browser' should refer to a product (Chrome, Edge, etc.) and 'vendor' should refer to the organization responsible for one or more browsers (Google, Microsoft, etc.). Some vendors have had multiple browsers in the past.

lib/confluence/aggressive_removal.es6.js
```diff
@@ -43,97 +39,110 @@ foam.CLASS({
...
+          this.browserDAO.where(this.mlang.AND(
+              this.mlang.NEQ(this.Browser.BROWSER_NAME, browser.browserName),
+              this.mlang.IN(this.Browser.BROWSER_KEY, browserKeys)))
+                  .orderBy(this.mlang.DESC(this.Browser.RELEASE_DATE))
+                  .select(),
+          // Browsers released before the grace period "browser" vendor.
```
arobins: suggestion: ...by browser's vendor

lib/confluence/aggressive_removal.es6.js
```diff
@@ -43,97 +39,110 @@ foam.CLASS({
...
+      name: 'computeFromBrowsers',
+      documentation: `Compute aggressive removal value for "browser", given
+          arrays of "current" browsers from other vendors and "previous"
+          browsers (from "browser" vendor) that are older than the grace
+          period.
+
+          Let browserAPIs = <"browser" interfaces>
+          Let curAPIs = INTERSECT(<"current" interfaces>)
+          Let cutOffAPIs = <interfaces from newest releas in "previous">
+          Let oldAPIs = UNION(<"previous" releases other than "cutOff" release>)
```
arobins: interfaces from "previous"...

nit: cutOff isn't described in the docs.

lib/confluence/aggressive_removal.es6.js
```diff
@@ -43,97 +39,110 @@ foam.CLASS({
...
+            this.mlang.INTERSECT(
+                currentLessOne.map(brwsr => brwsr.interfaces),
+                // Intersect with removed-before-grace-period, not-added-back:
+                // allAPIsOlderThanPostRemoval \ postRemoval \ browser.
+                this.mlang.INTERSECT(
+                    this.mlang.UNION.apply(
+                        this.mlang,
+                        prevLessOne.map(brwsr => brwsr.interfaces)),
+                    this.mlang.SET_MINUS(
+                        postRemoval.interfaces,
```
arobins: SET_MINUS is nary now, right?

lib/confluence/aggressive_removal.es6.js
mdittmer: Punted to future commit. Filed #58

lib/confluence/aggressive_removal.es6.js
mdittmer: I suppose this should be better documented: `compute()` is only ever invoked from super class's `run()`, which provides the latest release of each browser as the `browsers` parameter.

Should I just add a comment here? `documentation:` for `compute()` call this out more explicitly?

lib/confluence/aggressive_removal.es6.js
mdittmer: I'll fix this when I do the big rename ( #58 ).

lib/confluence/aggressive_removal.es6.js
```diff
@@ -43,97 +39,110 @@ foam.CLASS({
...
-            });
-          }));
-        }
-        return Promise.all(computeForBrowsers);
+        // Select from first current browser.
+        return firstCurrent.interfaces.select(
+            // Intersect with other current browsers to get APIs shipped by
+            // all (non-"browser" vendor) current browsers.
+            this.mlang.INTERSECT(
+                currentLessOne.map(brwsr => brwsr.interfaces),
```
mdittmer: This was to avoid a name conflict with `browser` in the outer scope.

PR #61, opened 2017-05-10
Consolidate Release/API/Junction DAOs in contextualizing container; drop "mlang" property

mdittmer:
Many components access conceptually the same Release/API/Junction DAOs. Right now, they are manually injected/passed down between components. FOAM contexts were designed to avoid exactly this anti-pattern.

This PR consolidates the related DAOs into a component that exports them to its sub-context. Components that need access to these DAOs now import them (i.e., expect to be instantiated in a context that provides them).

Also, prefer `implements: [...Expressions]` to additional `mlang` property.


lib/confluence/api_velocity.es6.js
```diff
@@ -62,25 +37,18 @@ foam.CLASS({
...
-        return this.ExpressionsSingleton.create();
-      },
-    },
   ],
   methods: [
     {
       name: 'init',
       documentation: `The init function computes the API Velocity for
-          each major releases contained in the given releaseApiDAO and
+          each major releases contained in the given releaseWebInterfaceJunctionDAO and
```
arobins: Wrong DAO?

test/any/client/api_matrix-test.es6.js
```diff
@@ -4,81 +4,67 @@
...
-      'Array', 'find'));
+  };
+
+  beforeEach(function() {
+    Release = foam.lookup('org.chromium.apis.web.Release');
+    WebInterface = foam.lookup('org.chromium.apis.web.WebInterface');
+    Junction = foam.lookup('org.chromium.apis.web.ReleaseWebInterfaceJunction');
+    ApiMatrix = foam.lookup('org.chromium.apis.web.ApiMatrix');
+
+    let container = global.getReleaseApiContainer();
```
arobins: Better name might be createReleaseApiContainer, because it creates a new one every time.

lib/client/api_confluence.es6.js
```diff
@@ -11,18 +11,21 @@ require('../confluence/failure_to_ship_data.es6.js');
 foam.CLASS({
   name: 'ApiConfluence',
   package: 'org.chromium.apis.web',
+  implements: ['foam.mlang.Expressions'],
```
arobins: These changes seem like they could have been done in a separate PR.

lib/confluence/api_velocity.es6.js
mdittmer: No, this is a deliberate rename.

lib/client/api_confluence.es6.js
mdittmer: Could have... probably should have at least listed them in the PR description.

lib/confluence/api_velocity.es6.js
mdittmer: Talked offline. Nevermind. Need haz moar context :smirk_cat:

PR #62, opened 2017-05-10
Test and implement: API Velocity in terms of set ops

mdittmer:


lib/confluence/api_velocity.es6.js
```diff
@@ -38,88 +40,87 @@ foam.CLASS({
       },
     },
   ],
+
   methods: [
+    function run() {
```
arobins: Why does this not look like this other methods?

mdittmer: It was because I thought it was documented in a super class. (`function methodName() { ...}` is the shorthand method syntax when no other metadata required.) However, this isn't a `MetricComputer` *per se*; we probably should have a `Runnable` interface implemented by `MetricComputer` (and its subclasses) and by `ApiVelocity`. For now, I've copied the doc from `MetricComputer` in here.

PR #64, opened 2017-05-15
UnionDAO: union-of-none and resolve(null) fixes

mdittmer:
- The union of no DAOs is the empty read-only `NullDAO`;
- Fix NullDAO bug (and add test): When primary and secondary both `resolve(null)`, the `UnionDAO` was not calling `resolve(null)`.

### DIFF
--- a/lib/confluence/set_ops.es6.js
+++ b/lib/confluence/set_ops.es6.js
@@ -295,6 +295,7 @@ foam.CLASS({
 
   requires: [
     'foam.dao.NullDAO',
+    'foam.dao.ReadOnlyDAO',
     'org.chromium.dao.UnionDAO',
     'org.chromium.mlang.sink.AsyncSink',
     'org.chromium.mlang.sink.IntersectDecider',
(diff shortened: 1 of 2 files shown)


lib/confluence/set_ops.es6.js
```diff
@@ -252,9 +252,12 @@ foam.CLASS({
...
   listeners: [
     function onFind(data, o) {
-      if (data.errored === true || data.found !== null) return;
+      // Return if already reject()ed or resolve()d promise.
+      if (data.errored === true || data.found) return;
+      // Resolve with value of "o" if null previously found, or o is non-null.
+      var shouldResolve = data.found === null || o !== null;
       data.found = o;
-      if (o !== null) data.resolve(data.found);
+      if (shouldResolve) data.resolve(data.found = o);
```
arobins: Why the |= o| here? Did you mean to delete the previous line so that the assignment was only done when resolving the promise?

mdittmer: That's not right. Will fix.

PR #66, opened 2017-05-16
Test & implement: Failure to ship metric in terms of set ops

mdittmer:


lib/confluence/failure_to_ship.es6.js
```diff
@@ -41,92 +41,78 @@ foam.CLASS({
...
-              return this.failureToShipDAO.put(this.FailureToShipData.create(
-                {
+        return Promise.all(releases.map(
+            release => this.computeForRelease(releases, date, release)));
+      },
+    },
+    {
+      name: 'computeForRelease',
+      documentation: 'Compute failure to ship value for a single release.',
+      code: function(releases, date, release) {
```
arobins: This method looks very similar to the one I just reviewed. Maybe it should be shared?

mdittmer: Yes, I intend to consolidate these classes in a future PR. I filed #67 for this.

PR #70, opened 2017-06-05
Consolidate historical browser metric data

mdittmer:
Historical metric data (i.e., all metric data except "API Velocity" metric data) follow the same pattern. They capture a data point at a particular point in time, and store (1) the main browser release used for metric calculation, (2) any previous browser releases used for metric calculation, and (3) releases of other browsers used for metric calculation.

This information is now consolidated in a single class that also stores the "type" of metric associated with the data point (an enum value). Data attached to enum values capture metadata and documentation related to each individual metric.


lib/chart/browser_specific_chart.es6.js
```diff
@@ -30,7 +30,7 @@ function browserSpecificChartController($scope, $timeout) {
               },
               {
                 name: 'falure to ship',
-                value: d.numBrowserSpecific,
+                value: d.value,
```
arobins: Do you plan on merging the charts? They look even closer to identical now.

mdittmer: Good point. I hadn't thought much about it. I'll open an issue to that effect.

PR #77, opened 2017-06-15
Refactor ApiExtractor's postProcess to support multiple processors

mdittmer:
Initial processor implementations:

- Add APIs: For adding a known collection of APIs to an interface
- Remove Interfaces: For removing interfaces by name
- Copy to prototype: For copying APIs from an interface to its prototype interface


lib/web_catalog/post_processors.es6.js
```diff
@@ -0,0 +1,163 @@
...
+          name: 'apiCatalog',
+          typeName: 'JSON',
+        },
+        {
+          documentation: `The object graph.`,
+          name: 'og',
+          typeName: 'ObjectGraph',
+        },
+      ],
+      code: function(apiCatalog, og) {},
```
arobins: Should this throw so that we know that a postprocessor isn't properly defined?

lib/web_catalog/post_processors.es6.js
```diff
@@ -0,0 +1,163 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+'use strict';
+
+foam.INTERFACE({
+  name: 'ApiCatalogPostProcessor',
+  package: 'org.chromium.apis.web',
```
arobins: I think class-level docs for all of these would be a good idea.

lib/web_catalog/post_processors.es6.js
```diff
@@ -0,0 +1,163 @@
...
+  properties: [
+    {
+      class: 'String',
+      name: 'interfaceName',
+      documentation: `Name of interface to be extended.`,
+      required: true,
+    },
+    {
+      class: 'StringArray',
+      name: 'propertyNames',
```
arobins: Not all names will be properties, according to the class docs.

lib/web_catalog/post_processors.es6.js
```diff
@@ -0,0 +1,171 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+'use strict';
+
+foam.INTERFACE({
+  name: 'ApiCatalogPostProcessor',
+  package: 'org.chromium.apis.web',
+
+  documentation: `JSON<apiCatalog> mutators. Implementations perform
+      post-processing steps on an "apiCatalog" extracted ObjectGraph "og".
```
arobins: extracted from?

PR #81, opened 2017-06-21
gRPC-based DatastoreDAO (introduces dep on @google-cloud/datastore)

mdittmer:


test/gcloud/dao-integration.js
```diff
@@ -0,0 +1,318 @@
...
+              of: BatchedItem,
+              protocol: env.CDS_EMULATOR_PROTOCOL,
+              host: env.CDS_EMULATOR_HOST,
+              port: env.CDS_EMULATOR_PORT,
+              projectId: env.CDS_PROJECT_ID
+            }, ctx);
+        var promises = [];
+        for ( var i = 0; i < 1000; i++ ) {
+          promises.push(dao.put(BatchedItem.create({
+            id: (i).toString()
```
arobins: Are the brackets around i significant?

test/gcloud/dao-integration.js
```diff
@@ -0,0 +1,318 @@
...
+            of: cls,
+            projectId: env.CDS_PROJECT_ID
+          }, opt_ctx || ctx);
+    });
+  };
+
+  // From foam2's test/helpers/generic_dao.js.
+  global.genericDAOTestBattery(daoFactory);
+
+  describe('construction', function() {
```
arobins: How much of the stuff below is testing the gRPC implementation? Testing the base class would probably be better done closer to the base class.

test/third_party/install.sh
```diff
@@ -0,0 +1,42 @@
...
+GCLOUD_SDK_TGZ="$BASE_DIR/google-cloud-sdk.tar.gz"
+GCLOUD_PATH="$(which gcloud)"
+
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+RED='\033[0;31m'
+NC='\033[0m'
+
+function win() {
+    printf "\n${GREEN}$1${NC}\n"
```
arobins: You could use $@ instead of $1. That will cause all arguments to be output instead of just the first.
You aren't using formatting, so echo would be more appropriate.
Having a leading newline in output feels odd.

mdittmer:
Currently no intention to merge this. Closing for now.

PR #83, opened 2017-07-17
Fix API service bugs: Missing require() and access constants from cla…

mdittmer:
…ss instead of instantiating invalid DAO container


mdittmer:
TBR @arobins

PR #85, opened 2017-07-17
Fix datastore query structure (and destructuring)

mdittmer:
Issue APIMatrix queries directly over relases IDs to avoid join query with
huge IN(AnArrayProperty, <way-too-many-values>). Support modest collection of
values on IN() clause with NoDisjunctionDAO wrapping DatastoreDAO.

### DIFF
--- a/lib/client/api_matrix.es6.js
+++ b/lib/client/api_matrix.es6.js
@@ -101,29 +101,27 @@ foam.CLASS({
             .select(this.AnonymousSink.create({
               sink: {
                 put: junction => {
-                  var apiKeys = junction.targetId.split('#');
-                  foam.assert(apiKeys.length === 2,
-                              `Expected API key of the form:
-                                  Iface#propOrMethod`);
-                  const k0 = apiKeys[0];
-                  const k1 = apiKeys[1];
-                  const k2 = junction.sourceId;
-                  // Function and Object's built-in methods can be
-                  // interface name or API name.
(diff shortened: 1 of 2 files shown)


lib/client/api_matrix.es6.js
```diff
@@ -87,24 +88,26 @@ foam.CLASS({
...
-                sink: {put: iface => {
-                  let k0 = iface.interfaceName;
-                  let k1 = iface.apiName;
-                  let k2 = releaseId;
+
+        return this.releaseWebInterfaceJunctionDAO.where(query)
+            .select(this.AnonymousSink.create({
+              sink: {
+                put: junction => {
+                  var apiKeys = junction.targetId.split('#');
```
arobins: This does not seem like a good way of getting the data that this method needs.

lib/client/api_matrix.es6.js
```diff
@@ -87,24 +88,26 @@ foam.CLASS({
...
+                  foam.assert(apiKeys.length === 2,
+                              `Expected API key of the form:
+                                  Iface#propOrMethod`);
```
arobins: The string "#" passes this check, so...you expect that the string have exactly one '#' character.

lib/client/api_matrix.es6.js
mdittmer: You mean, I should `.find()` the object itself, then extract the two pieces of metadata? Or just that there should be more substantial in-situ verification?

PR #90, opened 2017-08-03
Versioned Datastore import

mdittmer:
Import Datastore as versioned data using SyncDAOs. This change also modifies DatastoreContainer behaviour to correctly manage SyncDAOs in the data upload case and the web service case. It cannot be merged to master until all code has been ported to use versioned data.


main/datastore_update.es6.js
```diff
@@ -15,77 +14,337 @@ const path = require('path');
...
+            promises.push(dstDAO.put(array[i]));
+          }
+          return Promise.all(promises).then(function() {
+            logger.info(`Pushed ${array.length} ${srcDAO.of.id} to Datastore`);
+          });
+        });
+  }
+  function removeData(srcDAO, cmpDAO, dstDAO) {
+    const sink = foam.dao.ArraySink.create();
+    return cmpDAO.select(E.SET_MINUS(srcDAO, sink))
```
arobins: I'd probably add a check that prevents deletion of the entire database because nothing was loaded, and maybe because everything was marked as not in the loaded data.

mdittmer: Do you mean do a `select(COUNT())` on `srcDAO` to make sure it's non-empty?

arobins: That seems reasonable.

PR #93, opened 2017-08-29
Web service: Load local cache from disk, then polling sync from Datastore

mdittmer:
This change allows web service instances to serve from Datastore (slow) immediately, and load a local cache of data (and serve that, instead) in ~1min

**Summary of changes**

- `JDAO` (journaling DAO) data files from latest Datastore snapshot
- Scripts for updating (and testing load time of) journaled data on local disk from Datastore
- Store caches in forked processes to maintain HTTP service availability
- Use `BaseDatastoreContainer` used in forked processes (no need for container for all DAOs; only one per process); `DatastoreContainer` in main process tracking all DAOs
- Custom sync DAOs know where to find local journals and expose `hasSynced()` wired to base `SyncDAO`'s `synced` Promise
- `main/serve.js` exposes nothing but data API for now


lib/datastore/datastore_container.es6.js
```diff
@@ -240,6 +152,139 @@ foam.CLASS({
...
+            data: this.ConfluenceSyncDAO.create({
+              of: versionedCls,
+              remoteDAO: baseRemoteDAO,
+            })
+          })),
+        }, ctx);
+
+        this.info(`Serving read-only DAO of ${versionedCls.id} from Datastore`);
+        sync.hasSynced().then(function() {
+          proxy.delegate = sync;
```
arobins: Is this potentially racy, or does the mostly single-threaded nature of JS save you from having this switched out while something is using it?

main/forkScript.js
```diff
@@ -0,0 +1,67 @@
...
+const logger = foam.log.ConsoleLogger.create();
+const daoContainer = pkg.BaseDatastoreContainer.create({
+  gcloudAuthEmail: credentials.client_email,
+  gcloudAuthPrivateKey: credentials.private_key,
+  gcloudProjectId: credentials.project_id,
+  logger: logger,
+});
+const ctx = daoContainer.ctx;
+
+ctx.socketService.listening$.sub(function(sub, _, __, slot) {
```
arobins: Instead of copy/pasting this blob of code everywhere, maybe the foam2 repo should have a function that does this work on a provided context?

main/forkScript.js
```diff
@@ -0,0 +1,67 @@
...
+  if (!slot.get()) return;
+
+  sub.detach();
+  var stdin = require('process').stdin;
+  var buf = '';
+  stdin.on('data', function(data) {
+    buf += data.toString();
+  });
+  stdin.on('end', function() {
+    // TODO(markdittmer): Use secure parser.
```
arobins: I'd recommend at least having some sanity checks on this, like that the object ends up being an instance of something sane. Otherwise this feels like eval'ing from stdin, with a few extra steps.

lib/datastore/datastore_container.es6.js
mdittmer: This should be fine. In theory, a `ProxyDAO` implementation could asynchronously access its delegate twice under the assumption that it hasn't changed, but that would be a bug in the proxy's implementation. `ReadOnlyDAO` does not do this.

main/forkScript.js
mdittmer: I was trying to think through how that would work. I guess an argument or environment variable would have to be provided to the top-level script, that tells it what module to require before setting up the socket connection.

Is that what you had in mind?

main/forkScript.js
mdittmer: That's a good point. The verifier should, perhaps, come from the "provided context" you mention above.

PR #95, opened 2017-09-01
Refactor page + worker threads on UI

mdittmer:
**Summary of changes**

- New `ApiMatrixController` provides out-of-thread API to `api_catalog` and `catalog_table` Angular controller/component

- `ApiConfluence` moved out-of-thread (from page to worker on client)

- Use caching DAOs, don't bother syncing (i.e., versioning) on client; hence server API has `AdapterDAO` to provide unversioned data

- Attempt to abort `ApiMatrix.toMatrix` if new inputs (UI or data) are available; this doesn't work quite as well as we might hope because the DAO being `select()`ed over is asynchronous; the best we can do is stop doing work on each record and let the iteration finish

- BUG FIX: typo in `ApiMatrix`: `searchKey` not `seachKey`


lib/client/api_confluence.es6.js
```diff
@@ -66,7 +66,7 @@ foam.CLASS({
       documentation: `Read from apiVelocityDAO and returns API velocity
           metric for each browser release.`,
       returns: {
-        typeName: 'JSON',
+        typeName: 'Promise',
         documentation: `JSON of the form:
```
arobins: A promise for JSON of the form?

lib/client/api_matrix.es6.js
```diff
@@ -96,35 +114,86 @@ foam.CLASS({
...
+                                return;
+                              }
+
+                              const k0 = iface.interfaceName;
+                              const k1 = iface.apiName;
+                              const k2 = junction.sourceId;
+                              // Function and Object's built-in methods can be
+                              // interface name or API name.
+                              // So matrix[k0] is not sufficient to check if
+                              // k0 exits in matrix, use hasOwnProperty instead.
```
arobins: nit: k0 exists

lib/client/api_matrix.es6.js
```diff
@@ -96,35 +114,86 @@ foam.CLASS({
...
+                              if (!Object.prototype.hasOwnProperty.call(
+                                  matrix[k0], k1)) {
+                                matrix[k0][k1] = {};
+                              }
+                              matrix[k0][k1][k2] = true;
+                            }));
+                    },
+                  },
+                })).then(() => {
+                  if (detached) return null;
```
arobins: I don't know if there's a nice way of fixing it, but these five lines are repeated several times here.

lib/client/api_matrix.es6.js
```diff
@@ -387,3 +456,160 @@ foam.CLASS({
...
+      name: 'apiMatrix_',
+      factory: function() { return this.ApiMatrix.create(); },
+    },
+    {
+      name: 'matrix_',
+      factory: function() { return {}; },
+    },
+    {
+      name: 'matrixPromise_',
+      value: Promise.resolve({}),
```
arobins: Is this a good default? Wouldn't it be better to reject until a fetch has started?

lib/client/api_service.es6.js
```diff
@@ -11,90 +11,51 @@ require('../web_apis/release_interface_relationship.es6.js');
...
+  const workerEvents = foam.core.FObject.create();
+  const workerEventsBox = foam.box.EventDispatchBox.create({
+    target: workerEvents,
+  }, ctx);
+  registry.register('events', null, workerEventsBox);
+
+  const workerMessagePortBox = foam.box.MessagePortBox.create({
+      target: new Worker('worker.bundle.js'),
+  }, ctx);
+  workerMessagePortBox.delegate;
```
arobins: What is this doing?

lib/client/api_service.es6.js
```diff
@@ -11,90 +11,51 @@ require('../web_apis/release_interface_relationship.es6.js');
...
+  foam.box.NamedBox.create({
+    name: '/worker',
+    delegate: workerMessagePortBox,
+  }, ctx);
+
```
arobins: nit: remove extra newlines

lib/client/events.es6.js
```diff
@@ -0,0 +1,20 @@
...
+'use strict';
+
+foam.CLASS({
+  package: 'org.chromium.apis.web',
+  name: 'Event',
+});
+
+foam.CLASS({
+  package: 'org.chromium.apis.web',
+  name: 'NewMatrixEvent',
```
arobins: Is this supposed to extend Event?

lib/component/catalog_table.es6.js
```diff
@@ -27,6 +32,29 @@ function catalogTableController($scope, api) {
     $scope.$apply();
   };
 
+  workerEvents.sub(function(_, event) {
+    if (pkg.NewMatrixEvent.isInstance(event)) displayMatrix(event.matrix);
+  });
+
```
arobins: nit: extra newlines

lib/component/catalog_table.es6.js
```diff
@@ -5,20 +5,25 @@
...
 function catalogTableController($scope, api) {
   $scope.apiCatalogMatrix = {};
   $scope.showRows = {};
   $scope.currentPage = 0;
   $scope.gap = 5;
   $scope.itemPerPage = 15;
-  let ctrl = this;
-  let apiMatrix = api.matrix;
-  let setPageLength = function() {
+  const ctlr = this;
```
arobins: I'd go with controller. ctlr looks like a typo.

lib/component/catalog_table.es6.js
```diff
@@ -87,29 +105,27 @@ function catalogTableController($scope, api) {
...
-    let data = encodeURI(csv);
-    let link = document.createElement('a');
-    link.setAttribute('href', data);
-    link.setAttribute('download', filename);
-    link.click();
+    let releaseKeys = getReleaseKeys(ctlr.releases);
+    matrixCtlr.getMatrixCSV(releaseKeys).then(csv => {
+      if (csv === null) return;
+      if (!csv.match(/^data:text\/csv/i)) {
+        csv = 'data:text/csv;charset=utf-8,' + csv;
```
arobins: Are there usages of getMatrixCSV that don't want this prefix?

mdittmer:
FYI: Dependent FOAM2 PRs:

https://github.com/foam-framework/foam2/pull/619 (journal updates only)
https://github.com/foam-framework/foam2/pull/634 (for modeled return type metadata)
https://github.com/foam-framework/foam2/pull/635 (for serving default enum value predicates correctly)

lib/client/api_confluence.es6.js
mdittmer: Done.

lib/client/api_matrix.es6.js
mdittmer: Done.

lib/client/api_matrix.es6.js
mdittmer: My design thinking here was that any view requesting "the current matrix" always gets a valid matrix. Until data are loaded, the current matrix is `{}`.

lib/client/api_service.es6.js
mdittmer: I should add a note/TODO here, and perhaps open a FOAM2 issue. This tickles a lazy factory that starts the page/worker `MessagePort` handshake (even though there are no data to send over it yet). That way the handshake cost is not incurred when we are trying to deliver data for the first time.

mdittmer: TODO added.

lib/client/api_service.es6.js
mdittmer: Done.

lib/client/events.es6.js
mdittmer: Yes! Done.

lib/component/catalog_table.es6.js
mdittmer: Done.

lib/component/catalog_table.es6.js
mdittmer: Done.

lib/component/catalog_table.es6.js
mdittmer: Not at the moment, but there may be in the future. We may expose the CSV data over a REST API.

lib/client/api_matrix.es6.js
mdittmer: I think this should be cleaned up. There are two things that should be done (IMHO):

1. Create a `maybeContinue` that can be `then()` between each step; it will `reject()` to stop flow control (so there will be a `catch()` at the end of the method).
2. We should have a sink that will fetch data in batches and support flow control between batches because the DAOs here are sometimes implemented synchronously (so we'll keep iterating over >1k records synchronously).

mdittmer: Updated `TODO` to punt above-mentioned strategy to a future PR.

lib/client/api_matrix.es6.js
arobins: So why is it a promise at all? Default to empty, swap in a new one when available.

lib/controller/api_catalog.es6.js
```diff
@@ -6,36 +6,33 @@
...
 
 angular.module('confluence')
   .controller('catalogController',
     ['$scope', 'api', function($scope, api) {
+    console.log('************************************** catalogController');
     // Activate dropdown and tabs.
     $('.add-release-dropdown').dropdown();
     $('ul#view-tabs').tabs();
-    let apiMatrix = api.matrix;
+    const matrixCtlr = api.apiMatrixController;
```
arobins: nit: matrixController

lib/client/api_matrix.es6.js
mdittmer: This allows us to write `this.matrixPromise_.then()` without having to write a guard like `if (this.matrixPromise_) ...`.

lib/controller/api_catalog.es6.js
mdittmer: Done.

PR #101, opened 2017-09-08
Multi-matrix load

mdittmer:
This changes the `ApiMatrixController` API to use a String `id` to identify the view requesting an API Matrix. This allows for per-view `toMatrix()` request aborts, and per-view `NewMatrixEvent`s.


mdittmer:
@arobins FYI: This PR needs test fixes and documentation. I'll add you to reviewers when it's ready.

lib/client/api_matrix.es6.js
```diff
@@ -75,124 +138,79 @@ foam.CLASS({
...
+        documentation: `org.chromium.apis.web.MatrixMetadata.matrix. JSON:
             {interfaceName: {<apiName>: {<ReleaseKey>: true, ...}...}...}...}
             This structure simplifies for displaying a nested table.`,
       },
-      code: function(releaseKeys, options) {
-        const id = this.runInParallel ? 0 : ++this.matrixId_;
-        return this.verifyReleaseKeys_(releaseKeys).then(() => {
-          if (id && this.matrixId_ !== id) return null;
+      code: function(metadata) {
+        const id = metadata.id;
```
arobins: Is it correct that this ID-passing stuff is mostly driven by sharing a single instance of this class instead of constructing one instance per view?

mdittmer: That's correct. I don't think having separate instances would be much simpler though. Since the `ApiMatrix` and `ApiMatrixController` both live in a worker, creating multiple instances would turn id-passing into service object creation, registration, then binding.

mdittmer: I see. Same controller, multiple `APIMatrix`es. Yeah, that would probably work.

lib/client/api_matrix.es6.js
```diff
@@ -75,124 +138,79 @@ foam.CLASS({
...
+        const releaseKeys = metadata.releaseKeys;
+        if (this.toMatrixRequests_[id]) this.toMatrixRequests_[id].detach();
+
+        // Setup control flow for toMatrix request.
+        const sub = this.toMatrixRequests_[id] = foam.core.FObject.create();
+        let cancelled = false;
+        let CANCELLED = {};
+        sub.onDetach(() => cancelled = true);
+        const maybe = f => value => {
```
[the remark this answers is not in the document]
mdittmer: I think you're right. There was a version of this PR that had more coupling between the controller and the matrix, but with that gone this stands on its own.

lib/client/api_matrix.es6.js
```diff
@@ -546,72 +541,74 @@ foam.CLASS({
...
-      },
+    function onDAOChanges() {
+      for (const matrixData of this.matrixData_) {
+        this.onChanges(matrixData);
+      }
+    },
+    function onChanges(matrixData) {
+      this.apiMatrix_.toMatrix(matrixData.metadata)
+          .then(matrix => {
+            matrixData.metadata.matrix = matrix;
```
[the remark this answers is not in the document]
mdittmer: Yes. Done.

mdittmer:
I'm closing this in light of #102. The strategy, for now, will be to open these views in a separate window (see #106).

PR #104, opened 2017-09-11
Update browser release history

mdittmer:
A few typos fixed and a new version record matching strategy:

**Default strategy**: Use official desktop release data for browser major
  version number.

**Safari strategy**: Locate exact WebKit entry in
  https://en.wikipedia.org/wiki/Safari_version_history; use earliest release
  date at or prior to this entry that (1) has the same OS version, and (2)
  contains non-trivial changes (i.e., more than "Security update", "Improves
  website compatibility, application stability", etc.).

**Edge strategy**: Use release date of exact EdgeHTML version number.


arobins:
I think you need to update a test that's (possibly incorrectly) dependent on this file.

PR #110, opened 2017-09-12
Implement a class whitelist for network and postMessage transport

mdittmer:
This change switches to using secure `foam.box.Context`s with a classlist
injected from `data/class_whitelist.json`.


mdittmer:
Forgot to mention, dependent FOAM2 PR: https://github.com/foam-framework/foam2/pull/658

PR #112, opened 2017-09-13
Fix class whitelist and arrange for sending less DAOs over sockets to child processes

mdittmer:
This uses a verbose means of class lookup and constructing objects in the right context in `sync_dao` that should be cleaned up.


lib/sync_dao.es6.js
```diff
@@ -93,22 +95,29 @@ foam.CLASS({
       factory: function() {
         this.validate();
 
+        // TODO(markdittmer): Clean this up. Need a way to override lookup and
```
mdittmer: @adamvy any insight on how to build axioms for the right pattern here? What I'm manually doing is the following:

- This instance may have been created in a whitelisted context. Now that deserialization is complete, it wishes to instantiate not-whitelisted-for-serialization objects.
- It does this by using another context (that contains important info, but isn't whitelist-managed) to perform `lookup()` (see `this.lookup()` implementation), and
- Passing in the context to each `create()` call

This gets the job done, but it's really verbose. Also, this class `requires`ing classes does not guarantee that the "other context I want to use for lookup and creation" knows about those classes. (It happens to work in this case, but such a guarantee would be nice.)

Any thoughts on how to reify this into a common pattern? I think it will come up again as more and more as server-side apps need isolated contexts.

mdittmer:
@arobins PTAL

lib/datastore/datastore_container.es6.js
```diff
@@ -160,114 +287,6 @@ foam.CLASS({
   ],
 });
 
-foam.ENUM({
```
arobins: Did this need to be moved? Could it have been a separate PR to make comparing old and new possible?

lib/sync_dao.es6.js
```diff
@@ -119,13 +128,30 @@ foam.CLASS({
...
     },
     {
       class: 'FObjectProperty',
       of: 'org.chromium.apis.web.SyncDAOMonitor',
       name: 'monitor_',
-      factory: function() { return this.SyncDAOMonitor.create(); },
+      transient: true,
+      factory: function() {
+        return foam.lookup('org.chromium.apis.web.SyncDAOMonitor')
+          .create(null, this.container.ctx);
```
arobins: nit: continuation indent is +4

lib/datastore/datastore_container.es6.js
mdittmer: I just wanted to keep it above its first use. Code should still work with it in its old place, though.

mdittmer: Done.

lib/sync_dao.es6.js
mdittmer: Done.

mdittmer:
@adamvy I'm merging this in, but I think we should consult about the issue raised in my inline comment.

PR #115, opened 2017-09-13
Friendly safari versions

mdittmer:
This is towards #4 with a static mapping. It resolves the problem at the UI level. Eventually, the object graph data collection and Datastore should both store Safari (not WebKit) version numbers.

### DIFF
--- a/lib/web_apis/release.es6.js
+++ b/lib/web_apis/release.es6.js
@@ -7,6 +7,39 @@ foam.CLASS({
   name: 'Release',
   package: 'org.chromium.apis.web',
   ids: ['releaseKey'],
+
+  constants: {
+    // HACK(markdittmer): A bunch of Safari data were collected before userAgent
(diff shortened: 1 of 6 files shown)


lib/web_apis/release.es6.js
```diff
@@ -24,6 +57,22 @@ foam.CLASS({
     },
     {
       class: 'String',
+      name: 'friendlyBrowserVersion',
+      transient: true,
+      getter: function() {
+        const names = this.FRIENDLY_BROWSER_VERSIONS;
+        if (!names[this.browserName]) return this.browserVersion;
+        const versionNames = names[this.browserName];
+        const version = this.browserVersion.split('.');
```
arobins: This works for the current state of data for Safari. Simple prefix matching might be more useful long-term, though I guess you said you wanted to do the right thing (deal with it as the data is gathered) so that long-term usage probably doesn't exist.

PR #129, opened 2017-09-18
Add MDAO index on SOURCE_ID as implementation detail of ConfluenceSyncDAO

mdittmer:
This yields substantial improvements in server-side junction DAO performance.

### DIFF
--- a/lib/sync_dao.es6.js
+++ b/lib/sync_dao.es6.js
@@ -2,6 +2,9 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+require('./web_apis/release_interface_relationship.es6.js');
+const pkg = org.chromium.apis.web;
+
 foam.CLASS({
   package: 'org.chromium.apis.web',
   name: 'SyncDAOMonitor',
@@ -85,6 +88,7 @@ foam.CLASS({
     'foam.dao.SyncDAO',
     'foam.dao.sync.VersionedSyncRecord',
     'org.chromium.apis.web.SyncDAOMonitor',
+    'org.chromium.apis.web.VersionedReleaseWebInterfaceJunction',
   ],
   imports: ['container'],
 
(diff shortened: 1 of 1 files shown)


lib/sync_dao.es6.js
```diff
@@ -97,15 +101,24 @@ foam.CLASS({
 
         // TODO(markdittmer): Clean this up. Need a way to override lookup and
         // requiresCreate.
-        var lookup = this.lookup.bind(this);
+        const lookup = this.lookup.bind(this);
+
+        const indexedMDAO = lookup('foam.dao.MDAO').create({
+          of: this.of,
+        }, this.container.ctx);
+        // Index releaes ID on release/interface junction DAO.
```
foolip: typo: releaes

lib/sync_dao.es6.js
```diff
@@ -97,15 +101,24 @@ foam.CLASS({
...
+        if (this.of === pkg.VersionedReleaseWebInterfaceJunction) {
```
arobins: Does the MDAO have any benefits (or drawbacks) if you don't hit this codepath? This feels like an OO antipattern, and the Right Way would usually be to subclass for this case.

lib/sync_dao.es6.js
```diff
@@ -132,7 +145,7 @@ foam.CLASS({
       factory: function() {
         // TODO(markdittmer): Clean this up. Need a way to override lookup and
         // requiresCreate.
-        var lookup = this.lookup.bind(this);
+        const lookup = this.lookup.bind(this);
```
arobins: Unrelated change?

lib/sync_dao.es6.js
mdittmer: Do you mean implement `VersionedReleaseWebInterfaceJunctionMDAO` with default `of` class and default property index setup?

lib/sync_dao.es6.js
mdittmer: Yeah. I guess.

mdittmer:
@arobins PTAL at response. I agree is-an-instance is a bit hacky, but I want to clarify the alternative.

lib/sync_dao.es6.js
arobins: Something like that

PR #130, opened 2017-09-18
Add front-end server caching at the request level

mdittmer:
This strategy is used for DAOs, which change infrequently (once per browser release). Caches are completely cleared whenever new data are sync from Datastore.


lib/server/cache_handler.es6.js
```diff
@@ -0,0 +1,272 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+foam.CLASS({
+  package: 'org.chromium.apis.web',
+  name: 'ClockCache',
```
arobins: Add documentation. Is this approximately an LRU cache?

mdittmer: It is [exactly that](https://en.wikipedia.org/wiki/Page_replacement_algorithm#Clock).

mdittmer: ... willl add documentation.

Issue #126, opened 2017-09-15
Add note about specs


RByers:
Somewhere in the description we should note that "API confluence" focuses on interoperability of implementations, which is only one piece of the story for overall web platform health.  In particular, whether an API is well defined in a spec  (and potentially what level of maturity that spec has reached and how well it's tested by web-platform-tests) is an important part of the question of whether a browser-specific API is a bad thing or not.  But comparing to specs is out of scope for this particular tool at this time.

RByers:
@foolip @drufball want to propose any specific text on the opening page for this?

foolip:
We have "This site is a work in progress. Data reported on the site may not be an accurate reflection of web platform implementations. Please refer to known issues on Github for more information." Add somewhere: "Status in standards is not part of the data or analysis, and there is no guarantee that a widely-supported API is interoperable in its details, or will remain part of the web platform."

Yes, yes, we normally shouldn't be removing widely-supported APIs, but very occasionally we have, and I don't think we regret all of them.

foolip:
@mdittmer, should this be closed now?

RByers:
What we have now addresses my concern.  Thanks!

PR #131, opened 2017-09-18
Add warning about standards missing from analysis

mdittmer:
Towards #126. @foolip PTAL.


foolip:
Where does this warning show up? I can't see it live.

PR #135, opened 2017-09-18
Custom server

mdittmer:
This implements a workaround for [the problem described here](https://github.com/GoogleChrome/confluence/pull/130/commits/866682f67a7c4613310188773ba6fde0d3edce2e). The purpose of doing so is the following:

Currently cached query results are sent to the client immediately :+1:. However, the query is still forwarded to the in-memory database that processes queries in series :-1:. This change allows for the former and not the latter.

The change also implements static file caching.


lib/server/file_handlers.es6.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+  ],
+
+  methods: [
+    function canHandleRequest(req) {
+      if (!this.dir) return false;
+
+      // Check the URL for the prefix.
+      const target = this.getPathTarget_(req.url);
+      if (target === null) return false;
```
arobins: return target !== null

PR #149, opened 2017-09-19
Update journaled data based on latest Cloud Datastore push

mdittmer:


mdittmer:
@foolip be not dismayed. We will get you more ink soon. :red_circle: :wavy_dash: :secret:

PR #151, opened 2017-09-19
Rework build process

mdittmer:
New process allows:

- Separate `dev` and `prod` webpack build configurations
- Minification (to degrees)
- Chunking of vendor code (for now, just one big chunk)
- Worker loading that is compatible with chunking
- Deployment script that captures git hash of Confluence and FOAM2 in version

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -1,0 +1,27 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+'use strict';
+
+const merge = require('webpack-merge');
+
+const C = require('./webpack.constants.js');
+const common = require('./webpack.common.js');
+
+const execSync = require('child_process').execSync;
+execSync(`node '${C.FOAM_DIR}/tools/build.js'  ${C.FOAM_FLAGS} ${C.FOAM_BIN_TMP_PATH}`);
+
+module.exports = merge(common, {
+  devtool: 'inline-source-map',
+  module: {
+    rules: [
+      {
+        test: /\.es6\.js$/,
+        loader: 'babel-loader',
+        options: {
+          plugins: ['transform-runtime'],
+        },
+      },
+    ],
+  },
+});
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -9,7 +9,7 @@ const C = require('./webpack.constants.js');
 const common = require('./webpack.common.js');
 
 const execSync = require('child_process').execSync;
-execSync(`node '${C.FOAM_DIR}/tools/build.js'  ${C.FOAM_FLAGS} ${C.FOAM_BIN_TMP_PATH}`);
+execSync(`node '${C.FOAM_DIR}/tools/build.js'  ${C.FOAM_FLAGS} "${C.FOAM_BIN_TMP_PATH}"`);
 
 module.exports = merge(common, {
   devtool: 'inline-source-map',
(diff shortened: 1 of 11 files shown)


scripts/deploy.sh
```diff
@@ -0,0 +1,46 @@
...
+
+pushd "${WD}/.."
+CONFLUENCE_VERSION=$(git rev-parse HEAD)
+popd
+
+pushd "${WD}/../node_modules/foam2"
+FOAM2_VERSION=$(git rev-parse HEAD)
+popd
+
+if [ "${CONFLUENCE_VERSION}" == "${FOAM2_VERSION}" ]; then
```
arobins: This check fails to fail if the script is not executed from a git repo because empty strings are equal.

scripts/deploy.sh
```diff
@@ -0,0 +1,46 @@
...
+function warn() {
+    printf "\n${YELLOW}$1${NC}\n"
+}
+
+function error() {
+    printf "\n${RED}$1${NC}\n"
+}
+
+
+pushd "${WD}/.."
```
arobins: I usually pipe these to /dev/null. The output isn't very useful in the context of a script.

scripts/deploy.sh
```diff
@@ -0,0 +1,46 @@
...
+FOAM2_VERSION=$(git rev-parse HEAD)
+popd
+
+if [ "${CONFLUENCE_VERSION}" = "${FOAM2_VERSION}" ]; then
+  error "FOAM2 not not under version control"
+  exit 1
+fi
+win "Preparing build for Confluence@${CONFLUENCE_VERSION}, FOAM2@${FOAM2_VERSION}"
+
+webpack --config "${WD}/../config/webpack.prod.js"
```
arobins: `if ! webpack ...; then` is a more compact way of writing it, and is slightly less prone to being broken accidentally.

scripts/deploy.sh
```diff
@@ -0,0 +1,46 @@
...
+
+pushd "${WD}/.."
+CONFLUENCE_VERSION=$(git rev-parse HEAD)
+popd
+
+pushd "${WD}/../node_modules/foam2"
+FOAM2_VERSION=$(git rev-parse HEAD)
+popd
+
+if [ "${CONFLUENCE_VERSION}" = "${FOAM2_VERSION}" ]; then
```
arobins: zsh docs recommend using `[[ ]]`

config/webpack.dev.js
```diff
@@ -0,0 +1,27 @@
...
+// found in the LICENSE file.
+'use strict';
+
+const merge = require('webpack-merge');
+
+const C = require('./webpack.constants.js');
+const common = require('./webpack.common.js');
+
+const execSync = require('child_process').execSync;
+execSync(`node '${C.FOAM_DIR}/tools/build.js'  ${C.FOAM_FLAGS} ${C.FOAM_BIN_TMP_PATH}`);
```
arobins: Quote at least the temp path

scripts/deploy.sh
arobins: Oh, and this fails to fail if the script is executed from outside a git repo.

scripts/deploy.sh
```diff
@@ -0,0 +1,56 @@
...
+
+pushd "${WD}/../node_modules/foam2" > /dev/null
+FOAM2_VERSION=$(git rev-parse HEAD)
+popd > /dev/null
+
+if [[ "${CONFLUENCE_VERSION}" = "" ]]; then
+  error "Main source is not under version control. Is this not a git clone?"
+  exit 1
+fi
+if [[ "${CONFLUENCE_VERSION}" = "" || "${CONFLUENCE_VERSION}" = "${FOAM2_VERSION}" ]]; then
```
arobins: First check should be on FOAM_VERSION

PR #168, opened 2017-09-21
Fixes #165 - Changes overflow:hidden to be horizontal only.

### LINKED ISSUE #165: Doesn't scroll on Firefox or Firefox
Well this is embarrassing 😉. Our interop dashboard is unusable in Firefox because it doesn't scroll due to an `overflow:hidden` rule on body.  We should figure out why Chrome and Firefox behave differently in that regard (@bokand do you know?).  But separately let's just fix it ASAP.

karlcow:
The combination of overflow and defining size of the viewport will clip out the content and then block the scrolling on Firefox Android 57-

### DIFF
--- a/static/index.html
+++ b/static/index.html
@@ -26,7 +26,7 @@
         flex-direction: column;
         height: 100vh;
         width: 100vw;
-	overflow: hidden;
+	overflow-x: hidden;
       }
       main {
 	display: flex;


karlcow:
I signed it

karlcow:
Hold on on this pull request.
The discussion is still going on in https://github.com/webcompat/web-bugs/issues/10168 on the source of the issue.

karlcow:
This has been resolved by #176

PR #184, opened 2017-09-25
Fix generated links in Firefox

### LINKED ISSUE #181: Links to failure lists don't work in Firefox
STEPS TO REPRODUCE (in Firefox):

1. Load https://web-confluence.appspot.com/#!/confluence
2. Click "FAILURE TO SHIP".
3. Select any data point (the first data point at the "550" level for Safari, aka Safari 8.0, will do).
4. Click the little box-with-arrow after the number in the resulting popup.

EXPECTED RESULTS: Opens the actual list of things failing to ship in a new tab.  That's what it does in Chrome and Safari.  I can't test what it does with Edge, because in Edge I don't see the graph at all.

ACTUAL RESULTS: Nothing happens.

It's entirely possible that this is a Firefox issue, of course.  I'm happy to do some debugging, but it would be fairly helpful to know what the site is actually trying to do here, without having to reverse-engineer it from the minified source.

mdittmer:
This fixes #181

### DIFF
--- a/lib/component/catalog_table.es6.js
+++ b/lib/component/catalog_table.es6.js
@@ -198,8 +198,10 @@ function catalogTableController($scope, api) {
       }
       let data = encodeURI(csv);
       let link = document.createElement('a');
+      link.style.display = 'hidden';
       link.setAttribute('href', data);
       link.setAttribute('download', filename);
+      document.body.appenChild(link);
       link.click();
     });
   };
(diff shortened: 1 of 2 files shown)


lib/component/catalog_table.es6.js
```diff
@@ -198,8 +198,10 @@ function catalogTableController($scope, api) {
       }
       let data = encodeURI(csv);
       let link = document.createElement('a');
+      link.style.display = 'hidden';
```
foolip: Should this be display:none? hidden isn't a valid value I think.

foolip: Although, can `location.href` be used to navigate instead?

PR #191, opened 2017-09-25
Consolidate datastore update logic in set_ops and new class

mdittmer:
Also, test new class.

### DIFF
--- a/lib/datastore/updater.es6.js
+++ b/lib/datastore/updater.es6.js
@@ -19,10 +19,10 @@ foam.CLASS({
         this.info(`Unversioning ${array.length} ${versionedDAO.of.id} objects`);
         const cls = unversionedDAO.of;
         let promises = [];
-        for (let i = 0; i < array.length; i++) {
+        for (const item of array) {
           // Do not unversion deleted records.
-          if (!array[i].deleted_)
-            promises.push(unversionedDAO.put(cls.create(array[i])));
+          if (!item.deleted_)
+            promises.push(unversionedDAO.put(cls.create(item)));
         }
         return Promise.all(promises).then(() => {
           this.info(`Unversioned ${array.length} ${versionedDAO.of.id} objects`);
@@ -45,9 +45,9 @@ foam.CLASS({
               const array = sink.array;
               this.info(`Pushing ${array.length} ${srcDAO.of.id} to Datastore`);
               let promises = [];
-              for (var i = 0; i < array.length; i++) {
+              for (const item of array) {
                 // Push versioned class object to SyncDAO: dstDAO.
-                let datum = dstDAO.of.create(array[i]);
+                let datum = dstDAO.of.create(item);
                 promises.push(dstDAO.put(datum));
               }
               return Promise.all(promises).then(() => {
@@ -71,8 +71,8 @@ foam.CLASS({
         const array = sink.array;
         this.info(`Deleting ${array.length} ${srcDAO.of.id} from Datastore`);
         let promises = [];
-        for (var i = 0; i < array.length; i++) {
-          promises.push(dstDAO.remove(array[i]));
+        for (const item of array) {
+          promises.push(dstDAO.remove(item));
         }
         return Promise.all(promises).then(() => {
           this.info(`Deleted ${array.length} ${srcDAO.of.id} from Datastore`);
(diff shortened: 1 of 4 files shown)


lib/datastore/updater.es6.js
```diff
@@ -0,0 +1,83 @@
...
+
+  imports: ['info'],
+
+  methods: [
+    function unversionData(versionedDAO, unversionedDAO) {
+      return versionedDAO.select().then(sink => {
+        const array = sink.array;
+        this.info(`Unversioning ${array.length} ${versionedDAO.of.id} objects`);
+        const cls = unversionedDAO.of;
+        let promises = [];
```
foolip: Heh, this is exactly like some code I wrote just yesterday. I used const here though, but OK to use let if you have a convention about mutability that uses let even when const would work.

lib/datastore/updater.es6.js
```diff
@@ -0,0 +1,83 @@
...
+        for (let i = 0; i < array.length; i++) {
```
foolip: This can be a a for-of loop.

lib/datastore/updater.es6.js
```diff
@@ -0,0 +1,83 @@
...
+              });
+            });
+    },
+    function removeData_(srcDAO, cmpDAO, dstDAO) {
+      const sink = foam.dao.ArraySink.create();
+
+      // Quick sanity check against deleting entire database.
+      return srcDAO.select(this.COUNT()).then(countSink => {
+        foam.assert(countSink.value > 0,
+                    'Datastore update should not empty entire database' +
```
foolip: Sounds like a good rule of thumb :)

test/any/datastore/updater-integration.es6.js
```diff
@@ -0,0 +1,182 @@
...
+      syncRecordDAO: MDAO.create({of: VersionedSyncRecord}),
+      remoteDAO: VersionNoDAO.create({
+        delegate: datastoreDAO,
+      }),
+      polling: true,
+    });
+    updater = DatastoreUpdater.create();
+  });
+
+  it('should only perform necessary put/remove updates', done => {
```
foolip: What is `it` and why does it have a callback instead of a promise?

test/any/datastore/updater-integration.es6.js
```diff
@@ -0,0 +1,182 @@
...
+        const intervalId = setInterval(() => {
+          if (dstOpsStarted === dstOpsFinished) {
+            clearInterval(intervalId);
+            resolve();
+          }
+        }, 1);
+      });
+    }
+
+    dstPut(VersionedItem.create({
```
foolip: From here on I don't quite know what I'm looking at. Can you add a comment?

test/any/datastore/updater-integration.es6.js
```diff
@@ -0,0 +1,182 @@
...
+          id: 7,
+          data: 'toBeStillDeleted7',
+        }));
+        return dstDAO.sync().then(() => dstWait());
+      }).then(() => datastoreDAO.find(3))
+      .then(same3 => same3version = same3.version_)
+      .then(() => datastoreDAO.find(7))
+      .then(deleted7 => {
+        deleted7version = deleted7.version_;
+        expect(deleted7.deleted_).toBe(true);
```
foolip: Hmm, is this actually a test? Are the tests run automatically by Travis? I was about to comment that this was a lot of code with no tests, but maybe this is it?

test/any/datastore/updater-integration.es6.js
mdittmer: `it` is JasmineJS's test spec API. It may have a promisey API by now, but I'm used to depending on a version of JasmineJS that uses "you specify a single 'done' argument" as the API for async tests.

test/any/datastore/updater-integration.es6.js
mdittmer: Yes. All the `expect` calls in this file constitute the test expectations. These tests run on Travis (they were breaking before my follow-up commit!).

lib/datastore/updater.es6.js
mdittmer: Heh. Yeah. It's saved em before. This is actually copied from `main/datastore_update.es6.js`.

lib/datastore/updater.es6.js
mdittmer: Yeah. I tend to use let if I *know* I intend to mutate *something* within the object (even though that's not what ES2015 enforces :roll_eyes:)

foolip: That's fine, I'll know to expect just that going forward.

mdittmer:
Filed #196

PR #192, opened 2017-09-25
Split update logic into multiple phases

mdittmer:
OG => Unversioned Journal => Datastore + Versioned Journal


mdittmer:
Mistaken dup of #191 Closing

PR #193, opened 2017-09-25
Split update logic into multiple phases

mdittmer:
OG => Unversioned Journal => Datastore + Versioned Journal

This will obsolete `main/datastore_update.es6.js`. #191 must land first.


main/journal_to_datastore_and_journal.es6.js
```diff
@@ -0,0 +1,233 @@
...
+
+function getJournalDAO(name, cls, ctx, mode) {
+  const filename = path.resolve(__dirname, `../data/${name}-journal.js`);
+  logger.info(`Creating JDAO (mode=${mode}) in ${filename}`);
+  return foam.dao.JDAO.create({
+    of: cls,
+    delegate: foam.dao.MDAO.create({of: cls}, ctx),
+    journal: foam.dao.NodeFileJournal.create({
+      of: cls,
+      fd: fs.openSync(
```
foolip: Line wrapping here is weird.

main/journal_to_datastore_and_journal.es6.js
```diff
@@ -0,0 +1,233 @@
...
+// In-memory DAOs of new data to be imported.
+const releaseImportDAO = importCtx.releaseDAO;
+const webInterfaceImportDAO = importCtx.webInterfaceDAO;
+const releaseWebInterfaceJunctionImportDAO =
+    importCtx.releaseWebInterfaceJunctionDAO;
+const browserMetricsImportDAO = importCtx.browserMetricsDAO;
+const apiVelocityImportDAO = importCtx.apiVelocityDAO;
+
+//
+// Generic algorithm for data import:
```
foolip: This sounds less than trivial. Can't we write all the data to a new thing, and then point the old thing at the new thing? (Yay for comments, without it I probably wouldn't have known what I was looking at.)

main/og_to_journal.es6.js
```diff
@@ -0,0 +1,107 @@
...
+
+function getJournalDAO(name, cls, ctx, mode) {
+  const filename = path.resolve(__dirname, `../data/${name}-journal.js`);
+  logger.info(`Creating JDAO (mode=${mode}) in ${filename}`);
+  return foam.dao.JDAO.create({
+    of: cls,
+    delegate: foam.dao.MDAO.create({of: cls}, ctx),
+    journal: foam.dao.NodeFileJournal.create({
+      of: cls,
+      fd: fs.openSync(
```
foolip: A bit of deja vu here and the filename bit above. Is there some code sharing that should be done here?

main/journal_to_datastore_and_journal.es6.js
mdittmer: Datastore generally uses partitions for that. There's an outstanding bug to investigate why FOAM's Datastore/REST implementation does not handle partitions properly. It's a particularly frustrating problem because our tests against Datastore's emulator suggests that partitions ARE working properly, but the don't work in production.

main/og_to_journal.es6.js
mdittmer: Perhaps. I've been a little lazy about code sharing in `main` because we are likely to move to table-based (instead of join-based) model in our long term vision for performance improvements.

main/journal_to_datastore_and_journal.es6.js
foolip: OK, sounds like that is a bigger project in any event, not today.

main/og_to_journal.es6.js
foolip: That's OK, you're the maintainer and make the trade-off better than anyone else.

mdittmer:
Filed #196

PR #194, opened 2017-09-25
Add post-processor: ignore hyphenated CSSStyleDeclaration attributes

mdittmer:
This is towards #174


lib/web_catalog/post_processors.es6.js
```diff
@@ -85,6 +97,38 @@ foam.CLASS({
...
+      required: true,
+    },
+  ],
+
+  methods: [
+    function postProcess(apiCatalog, og) {
+      this.validate();
+
+      const apiNameRegExp = this.apiNameRegExp;
+      for (let i = 0; i < this.interfaceNames.length; i++) {
```
foolip: Can use for-of

lib/web_catalog/post_processors.es6.js
```diff
@@ -85,6 +97,38 @@ foam.CLASS({
...
+        let iface = this.interfaceNames[i];
+        if (apiCatalog[iface]) {
+          apiCatalog[iface] = apiCatalog[iface]
+              .filter(apiName => apiName.match(apiNameRegExp) !== null);
```
foolip: `apiNameRegExp.test(apiName)` would be shorter and returns a bool

PR #198, opened 2017-09-25
More default values to combat bad SyncDAO.remove()/DatastoreDAO.put() interaction

mdittmer:


mdittmer:
@arobins PTAL. Same rationale ase #190

PR #207, opened 2017-09-26
Delete version-controlled journaled caches

mdittmer:
`git lfs` quotas on GitHub are causing problems. These are to be hosted on
Cloud Storage now.


mdittmer:
@arobins @foolip this should fix the build by eliminating our dependency on `git lfs`. #206 allows us to still deploy to production based on ToT (once it lands).

PR #213, opened 2017-09-28
Unversioned journal => versioned journal based on READ from Datastore

mdittmer:
Get a versioned cache of data without pushing new cache to Datastore

### DIFF
--- a/main/journal_to_versioned_journal.es6.js
+++ b/main/journal_to_versioned_journal.es6.js
@@ -82,8 +82,10 @@ const importCtx = pkg.DAOContainer.create({
 
 // Context for reading from / writing to Datastore.
 //
-// TODO(markdittmer): Figure out why partitioning isn't working. This component
-// was originally to be instantiated in context:
+// TODO(markdittmer): Update after
+// https://github.com/foam-framework/foam2/issues/694 is fixed.
+//
+// This component was originally to be instantiated in context:
 //
 // foam.__context__.createSubContext({
 //   datastoreNamespaceId: 'someDatastoreNamespace'
--- a/main/journal_to_versioned_journal.es6.js
+++ b/main/journal_to_versioned_journal.es6.js
@@ -37,15 +37,15 @@ const ctx = foam.__context__.createSubContext({
   logger: logger,
 });
 
-function getJournalDAO(name, cls, ctx, mode) {
+function getJournalDAO(name, cls, ctx, flags) {
   const filename = path.resolve(__dirname, `../data/${name}-journal.js`);
-  logger.info(`Creating JDAO (mode=${mode}) in ${filename}`);
+  logger.info(`Creating JDAO (flags=${flags}) in ${filename}`);
   return foam.dao.JDAO.create({
     of: cls,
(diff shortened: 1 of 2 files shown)


main/journal_to_versioned_journal.es6.js
```diff
@@ -0,0 +1,246 @@
...
+  return Promise.all([
+    Promise.all(daosArray.map(sync)).then(() => {
+      return Promise.all(daosArray.map(daos => {
+        return updater.unversionData(daos.sync, daos.cache);
+      }));
+    }),
+    load(),
+  ]).then(() => {
+    return Promise.all(daosArray.map(function(daos) {
+      return updater.importData(daos.import, daos.cache, daos.export);
```
arobins: I think this only writes the delta between cache and import. The export DAO was truncated, so it will contain only the delta, not the whole set.

main/journal_to_versioned_journal.es6.js
```diff
@@ -0,0 +1,246 @@
...
+      pkg.ReleaseWebInterfaceJunction.id, pkg.ReleaseWebInterfaceJunction, ctx),
+  browserMetricsDAO: getReadJournalDAO(
+      pkg.BrowserMetricData.id, pkg.BrowserMetricData, ctx),
+  apiVelocityDAO: getReadJournalDAO(
+      pkg.ApiVelocityData.id, pkg.ApiVelocityData, ctx),
+}, ctx);
+
+// Context for reading from / writing to Datastore.
+//
+// TODO(markdittmer): Figure out why partitioning isn't working. This component
```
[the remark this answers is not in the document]
mdittmer: Filed https://github.com/foam-framework/foam2/issues/694

main/journal_to_versioned_journal.es6.js
```diff
@@ -0,0 +1,246 @@
...
+
+function getJournalDAO(name, cls, ctx, mode) {
+  const filename = path.resolve(__dirname, `../data/${name}-journal.js`);
+  logger.info(`Creating JDAO (mode=${mode}) in ${filename}`);
+  return foam.dao.JDAO.create({
+    of: cls,
+    delegate: foam.dao.MDAO.create({of: cls}, ctx),
+    journal: foam.dao.NodeFileJournal.create({
+      of: cls,
+      fd: fs.openSync(filename, mode),
```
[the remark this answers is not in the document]
mdittmer: Yes. That should be fine.

main/journal_to_versioned_journal.es6.js
mdittmer: Yes and no. The journals are truncated. However, they are installed as the delegate to the Datastore `SyncDAO` instances. As a result, waiting for `theSyncDAO.sync()` pushes the data currently in Datastore into the journal first, then `updater.importData()` appends the diff to the journal.

Unrelated: This does mean that data that are changed (rather than newly added) appear in the journal twice, and also means that we don't have a journal that contains *just* the changes that should later be pushed to Datastore.

mdittmer: Also, I extended the comment in the export DAO initialization to clarify how sync-then-add-updates works.

Issue #204, opened 2017-09-26
A bunch of incorrect aggressive removal listed for Safari 9.1.3 only


othermaciej:
I don't think any of the APIs below were actually removed in Safari 9.1.3, and the catalog seems to be aware that they were present both before and after.

https://web-confluence.appspot.com/#!/catalog?releaseKeys=%5B%22Safari_9.1.3_OSX_10.11.6%22,%22Safari_8.0.8_OSX_10.10.5%22,%22Safari_7.1.8_OSX_10.9.5%22,%22Safari_6.2.8_OSX_10.8.5%22,%22Safari_6.0.5_OSX_10.7.5%22,%22Safari_5.1.9_OSX_10.6.8%22,%22Chrome_57.0.2987.98_OSX_10.11.6%22,%22Firefox_52.0_OSX_10.11%22,%22Edge_14.14393_Windows_10.0%22%5D&releaseOptions=%7B%22Safari_9.1.3_OSX_10.11.6%22:false,%22Chrome_57.0.2987.98_OSX_10.11.6%22:true,%22Firefox_52.0_OSX_10.11%22:true,%22Edge_14.14393_Windows_10.0%22:true%7D&numAvailable=%5B4,5,6,7,8,9%5D&currentPage=0&gap=5&itemPerPage=15&searchKey=%22%22

RByers:
@mdittmer Maybe there's some stale data from a previous crawl there still?  I checked a couple and the API catalog does show them present in 8 and 10 but not 9 (eg. `CustomEvent.detail` and `window.indexedDB`).

Looks like just a few and the history isn't as important as the current state, so dropping urgent label.

RByers:
Poking at Safari 10.1.2 I see that properties which are supposed to be on `WebGLRenderingContext.prototype` are actually on an ancestor prototype which isn't exposed on window (`WebGLRenderingContextBasePrototype`).  But I don't see that interface listed in the catalog either.

@mdittmer we walk the whole prototype chain of interface objects we find, right?  Shouldn't we still find these properties somewhere, even if we end up reporting that they're on the wrong interface?

Re-adding urgent in case this represents a more fundamental data issue.

mdittmer:
I'm seeing the same thing. Everything other than `WebGLRenderingContext` APIs are indeed not present in Safari 9.1.

[...]

PR #217, opened 2017-09-28
ApiExtractor: Gather .prototype's prototypes' properties

mdittmer:
In the edge case of non-Function "libraries" that have a property named "prototype", gather properties from "prototype"'s prototype chain, as long as those properties don't appear to belong to some other interface.

### DIFF
--- a/lib/web_catalog/api_extractor.es6.js
+++ b/lib/web_catalog/api_extractor.es6.js
@@ -532,7 +532,7 @@ foam.CLASS({
                 const objectPrototypePropertyId = og.lookup('Object.prototype');
                 let prototypePropertyClassName =
                     this.getClassName_(prototypePropertyId, og);
-                while (prototypePropertyId &&
+                while (!og.isType(prototypePropertyId) &&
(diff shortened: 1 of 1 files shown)


mdittmer:
This is towards #204. In particular, it attempts to address "1." in https://github.com/GoogleChrome/confluence/issues/204#issuecomment-332542732.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -509,17 +509,37 @@ foam.CLASS({
...
               // objects can have prototype properties.
               let prototypePropertyProps = [];
               if (prototypePropertyId) {
-                prototypePropertyProps = this.getObjectProperties_(
-                    og, prototypePropertyId);
+                // Capture all APIs descending from <current object>.prototype
+                // that do not belong to a different interface.
+                //
+                // Walk the prototype chain of <current object>.prototype until:
+                // (1) the prototype's class name changes (i.e., prototype
```
arobins: Where is the check for a change in class name?

PR #220, opened 2017-10-10
New ApiExtractor

mdittmer:
Some of the peculiarities in the lists-of-APIs data proved near impossible to fix cleanly, prompting a rewrite of the `ApiExtractor`. This change includes the rewrite, a markdown file explaining the extraction algorithm, and associated `PostProcessor`/`test`/`data` changes.


CatalogDataCollection.md
```diff
@@ -0,0 +1,198 @@
...
+libraries, constructors, prototypes, and interface names. The following captures
+the overall algorithm, with reference to several heuristics that are described
+in more detail below.
+
+```
+ExtractApiCatalog(ObjectGraph):
+
+Let fns    := Map: ObjectGraph id -> [Strings]:
+                   All objects with a "prototype" property mapped to all names
+                   for this object based on "Function naming heuristics"
```
arobins: Can these be links?

lib/web_catalog/post_processors.es6.js
```diff
@@ -163,11 +163,48 @@ foam.CLASS({
...
+    function postProcess(apiCatalog, og) {
+      this.validate();
+
+      if (!apiCatalog.hasOwnProperty(this.fromInterfaceName))
+        return;
+
+      const fromArray = apiCatalog[this.fromInterfaceName];
+      const toArray = apiCatalog[this.toInterfaceName] || [];
+      apiCatalog[this.toInterfaceName] = toArray
+        .concat(fromArray.filter(api => !toArray.includes(api)));
```
arobins: nit: +2

lib/web_catalog/post_processors.es6.js
```diff
@@ -163,11 +163,48 @@ foam.CLASS({
 });
 
 foam.CLASS({
+  name: 'CopyToInterfaceProcessor',
+  package: 'org.chromium.apis.web',
+  implements: ['org.chromium.apis.web.ApiCatalogPostProcessor'],
+
+  documentation: 'Copy APIs from one interface to another.',
```
arobins: Document that things in fromInterfaceName will only be copied if they aren't already in toInterfaceName

lib/web_catalog/post_processors.es6.js
```diff
@@ -192,23 +229,53 @@ foam.CLASS({
...
 
-      var keyParts = og.getShortestKey(toProtoId).split('.');
-      foam.assert(
-          keyParts[0] === 'window' && keyParts[2] === 'prototype' &&
-              keyParts.length === 3,
-          `Expected CopyToPrototypeProcessor prototype to have string key of
-          window.<prototype ctor name>.prototype.`);
-      var toName = keyParts[1];
+      const interfaceNames = og.getKeys(toProtoId)
+            .filter(key => key.endsWith('.prototype'))
```
arobins: nit: -2

lib/web_catalog/post_processors.es6.js
```diff
@@ -192,23 +229,53 @@ foam.CLASS({
...
+              const parts = key.split('.');
+              return parts[parts.length - 2];
+            });
+      foam.assert(interfaceNames.length > 0,
+                  `Expected CopyToPrototypeProcessor prototype to have string
+                       key(s) of ...<prototype ctor name>.prototype.`);
 
-      apiCatalog[toName] = apiCatalog[toName]
-          .concat(apiCatalog[this.fromInterfaceName]);
+      const newAPIs = apiCatalog[this.fromInterfaceName];
```
arobins: The merge-with-shadowing logic here and above are the same. I don't know if it makes more sense to construct a CopyToInterfaceProcessor for each discovered interface and run it instead of duplicating the two lines, since the merge logic is unlikely to change.

CatalogDataCollection.md
mdittmer: Making these links is [painful/difficult inside markdown code blocks](https://stackoverflow.com/questions/20092433/placing-links-inside-markdown-code-blocks).

CatalogDataCollection.md
```diff
@@ -0,0 +1,198 @@
...
+
+Let protos := Map: ObjectGraph id -> [ObjectGraph ids]:
+                   All objects that are some "Foo.prototype" mapped to all such
```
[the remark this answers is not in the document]
mdittmer: The mapping is from id-of-`Foo.prototype` (and `Bar.prototype`) to array: `[Foo, Bar]`, when `Foo.prototype = Bar.prototype`.

lib/web_catalog/post_processors.es6.js
mdittmer: I'm not sure I follow. Do you mean construct a `CopyToInterfaceProcessor(foo, foo.__proto__)` instead of `CopyToPrototypeProcessor` at all?

arobins: No, I meant use CopyToInterfaceProcessor to do the copy here. This processor is a little bit more complicated because it scans for destinations first, but once it has them it's just repeating the copying operation as in CopyToInterfaceProcessor. But as I said, it avoids repeating about two lines of logic, so it might not make sense.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -5,49 +5,184 @@
...
+      value: true,
     },
     {
-      name: 'retainConstantMembers',
       class: 'Boolean',
-      documentation: `When retainConstantMembers is false
-        constant primitive properties are filtered.`,
-      value: false,
+      name: 'classNamesFromConstructorProperty',
+      documentation: `Deduce class names the "constructor" property. Not that
```
arobins: nit: Note that

lib/web_catalog/api_extractor.es6.js
```diff
@@ -5,49 +5,184 @@
...
+      documentation: `Deduce function names from object graph paths.`,
+      value: true,
     },
     {
-      name: 'retainConstantMembers',
       class: 'Boolean',
-      documentation: `When retainConstantMembers is false
-        constant primitive properties are filtered.`,
-      value: false,
+      name: 'classNamesFromConstructorProperty',
```
arobins: Design comment: I think for these optional heuristics, a better implementation might be to have a ClassNameHeuristic interface, implemented by ClassNameFromConstructorPropertyHeuristic, ClassNameFromGraphPathsHeuristic, etc. These can then be provided as appropriate, reordered, and potentially be unit tested.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -140,511 +242,428 @@ foam.CLASS({
...
+        this.functionDotPrototype_ = og.lookup('Function.prototype');
+
+        // Gather functions.
+        const allOgIds = og.getAllIds();
+        for (const id of allOgIds) {
+          if (this.isFunctionLike_(og, id)) {
+            const proto = og.lookup('prototype', id);
+            const names = this.getFunctionNames_(og, id);
+            let protos = this.protos_[proto] =
+                  this.protos_.hasOwnProperty(proto) ? this.protos_[proto] : [];
```
arobins: nit: less indentation

lib/web_catalog/api_extractor.es6.js
```diff
@@ -140,511 +242,428 @@ foam.CLASS({
...
-            this.extractAPI_(og, map, og.getPrototype(id),
-                             `${name}.__proto__`, visited, {proto: true});
+        }
+
+        // Gather "global libraries".
+        const root = og.getRoot();
+        const globalNames = og.getObjectKeys(root);
+        for (const name of globalNames) {
+          const id = og.lookup(name);
+          if ((!og.isType(id)) && (!this.isFunctionLike_(og, id))) {
```
arobins: I think there's probably a better way of expressing that the first loop should operate on some subset of the input, and that this loop should operate on the remainder. As it is now this check needs to be the negation of the check in the first, which may lead to bugs if it ever needs to be updated. Maybe a single loop that splits the input into 'things that should be processed like functions', 'things that should be processed as library objects', and 'other things', followed by loops over those subsets?

lib/web_catalog/api_extractor.es6.js
```diff
@@ -140,511 +242,428 @@ foam.CLASS({
...
+          const isLib = this.libs_.hasOwnProperty(id);
+          const isCtorProto = this.protos_.hasOwnProperty(id);
+
+          foam.assert(!((isCtor && isLib) || (isCtor && isCtorProto) ||
+                        (isLib && isCtorProto)),
+                      'ApiExtractor: Object with multiple personalities');
+
+          // Functions are interfaces, not prototypes. Just store properties and
+          // carry on.
+          if (isCtor) {
```
arobins: Since this will only ever be true for elements of this.fns_ (and the same for the other branches), why not split into multiple loops?

lib/web_catalog/api_extractor.es6.js
```diff
@@ -140,511 +242,428 @@ foam.CLASS({
...
+
+        // Copy APIs from instances down to closest interface prototype iff
+        // the API does not exist further down in the prototype chain.
+        for (const id of allOgIds) {
+          const isInterfacePrototype =
+              id => !og.isType(id) &&
+                  id !== this.objectDotPrototype_ &&
+                  id !== this.functionDotPrototype_;
+          const isCtor = this.fns_.hasOwnProperty(id);
+          const isCtorProto = this.protos_.hasOwnProperty(id);
```
arobins: Does this not need to filter out libraries? The documentation says it is supposed to.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -140,511 +242,428 @@ foam.CLASS({
...
+            // "Bar.prototype".
+            const ctorIds = this.protos_[id];
+            for (const ctorId of ctorIds) {
+              const ctorNames = this.getFunctionNames_(og, ctorId);
+              let protoId = id;
+              let protoNames = this.getClassNames_(og, protoId);
+              this.storeProtoAPIsForCtor_(og, ctorId, ctorNames, protoId,
+                                          protoNames);
+            }
+          } // else: This is an instance or a primitive value.
```
arobins: Sort of along the same line, adding an else branch that populates a list of non-constructor/library/prototype would mean less work later, because the filtering is done.

lib/web_catalog/api_extractor.es6.js
mdittmer: Good point. It actually is supposed to include libraries because sometimes properties on the global object are "interface-defining instances". They are filtered out during the second pass at instances to avoid stripping away library APIs that were (also) copied down to an interface.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -140,511 +242,434 @@ foam.CLASS({
...
+        // Gather "global libraries".
+        const root = og.getRoot();
+        const globalNames = og.getObjectKeys(root);
+        for (const name of globalNames) {
+          const id = og.lookup(name);
+          if ((!og.isType(id)) && (!this.isFunctionLike_(og, id))) {
+            const libId = id;
+            let names = this.getClassNames_(og, libId);
+            if (!names.includes(name)) names.push(name);
+            this.libs_[libId] = this.apisConcat_((this.libs_[libId] || []),
```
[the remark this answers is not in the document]
mdittmer: You mean have a method like `initLibId_ = libId => this.libs_[libId] = this.libs_[libId] || []`?

lib/web_catalog/api_extractor.es6.js
```diff
@@ -140,511 +242,434 @@ foam.CLASS({
...
+    function isFunctionLike_(og, id) {
+      return (!og.isType(id)) &&
+          og.getObjectKeys(id).includes('prototype');
+    },
+    function storeAPIsFromCtorOrLib_(og, ctorId) {
+      let apis = this.apis_[ctorId] || [];
+      // Do not copy properties belonging to Function API from ctors.
+      const newAPIs = this.apisFilter_(
+          apis, this.getClassAPIs_(og, ctorId).filter(api => ! [
+            'arguments',
```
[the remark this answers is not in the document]
mdittmer: The trouble with establishing it dynamically is that some browsers do not ship the whole API on `Function.prototype`. We could use a heuristic like "gather a handful of things that look like they're functions; use the intersection of their properties as the Function API", but that seems about as risky as a static list to me.

@arobins given the missing `Function.prototype` fields, WDYT is best? @foolip, do you have an opinion based on the stability of the `Function` interface?

lib/web_catalog/api_extractor.es6.js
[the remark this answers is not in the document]
arobins: I meant have apisConcat accept undefined as the first parameter, and have it default to an empty array.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -140,511 +242,440 @@ foam.CLASS({
...
+    function getInstanceAPIs_(id) {
+      const og = this.objectGraph;
+      // Exclude object-graph'ified properties "+reservedName+" and integers.
+      return og.getObjectKeys(id)
+          .filter(name => !/^[+].*[+]$/.test(name))
+          .filter(name => !/^[0-9]+$/.test(name));
+    },
+    function filterConstantAPIs_(id, name) {
+      const og = this.objectGraph;
+      if (this.constantTypes.length === 0) return true;
```
arobins: Is this needed? I guess it saves two OG calls, but the check at the end should work the same without this.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -140,511 +242,440 @@ foam.CLASS({
...
+
+      // Add to ctorId: New APIs that:
+      // (1) Are not exposed by some interface lower level than protoId's, AND
+      // (2) Are not already on ctorId.
+      const newAPIs = this.dedupFilter_(existingAPIs,
+                                       this.getClassAPIs_(protoId));
+      this.objectGraphApiMapper.map(ctorId, newAPIs, protoId);
+      this.apis_[ctorId] = ctorAPIs.concat(newAPIs);
+    },
+    function getAllProtosAPIs_(id) {
```
arobins: I think this produces the same list for an ID, without input from other parts of the processing. For heavily referenced prototypes, this approach will recompute the same list many times. Does it make sense to compute and store the flattened protoId to list of APIs once after all prototypes have been found?

lib/web_catalog/api_extractor.es6.js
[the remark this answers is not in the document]
arobins: Are you saying that there are cases where Function.prototype doesn't have all of the things in this list, but those things show up in prototypes that extend (is that what it's called?) Function.prototype?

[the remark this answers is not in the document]
foolip: Is the intent to exclude exactly this list of properties in all browsers, or to do exclude whatever is on Function.prototype in each browser? If it's the latter, then I'd be surprised if `Object.getOwnPropertyNames(Function.prototype)` isn't good enough in all browser versions we care about.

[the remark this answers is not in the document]
foolip: (Because it's such an old API.)

lib/web_catalog/api_extractor.es6.js
mdittmer: It's not functionally necessary. I added it as an optimization akin to the old `retainConstantMembers = false` case.

lib/web_catalog/api_extractor.es6.js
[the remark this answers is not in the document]
mdittmer: I'd have to go back and check, but I'm quite confident that some old browsers surface (some, if not all) of this API on instances only.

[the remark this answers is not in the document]
foolip: Ah. As discussed offline, I didn't realize this was a prototype vs. instance issue. And it seems fairly unlikely that more things will be added to all instances of Function, so leaving this along for now should be fine. Maybe a comment as a clue if anyone comes along debugging in the future.

lib/web_catalog/api_extractor.es6.js
arobins: bump

lib/web_catalog/api_extractor.es6.js
arobins: We discussed this offline and decided it wasn't worth doing.

PR #223, opened 2017-10-12
HttpJsonDAO and its container

mdittmer:
HttpJsonDAO is a DAO that will fetch an array of objects from a URL that
should return a JSON array.


mdittmer:
@arobins some context: @foolip and I were discussing ways to simplify our storage and caching strategy. Eventually, the code here should form a basis that eliminates the need for Cloud Datastore. We will load data from Cloud Storage URLs, and use custom health checks to avoid marking instances as "live" before they are finished loading the data. This will also eliminate the need for versioned data; we will either (1) deploy new code every time we update the data, or (2) use a polling strategy that simply checks a hash and replaces the whole cache with a fresh copy when it changes.

lib/http_json_dao_container.es6.js
```diff
@@ -0,0 +1,84 @@
...
+          unsafe: false,
+          classWhitelist: require('../data/class_whitelist.json'),
+        }, baseCtx));
+      },
+    },
+  ],
+
+  methods: [
+    function getDAO(cls) {
+      const ctx = this.ctx;
```
arobins: Something went wrong with the formatting in here

lib/http_json_dao_container.es6.js
```diff
@@ -0,0 +1,84 @@
...
+            delegate: this.ForkBox.create({
+              nodeParams: this.getForkNodeParams_(),
+              childScriptPath: this.getForkScriptPath_(),
+            }, ctx),
+            // Register skeleton over ConfluenceSyncDAO. ConfluenceClientDAO's
+            // delegate is box returned from register() call.
+          }, ctx).register(null, null, this.SkeletonBox.create({
+            data: this.ConfluencePollingDAO.create({
+              of: cls,
+              url: `https://storage.googleapis.com/web-api-confluence-data-cache/latest/json/${cls.id}.json`,
```
arobins: The URL prefix should probably be a constant at least. Possibly a configurable property.

lib/http_json_dao.es6.js
```diff
@@ -0,0 +1,173 @@
...
+      class: 'String',
+      name: 'url',
+      documentation: 'Configuration passed to delegate.',
+      required: true,
+    },
+    {
+      class: 'Array',
+      of: 'String',
+      name: 'safeProtocols',
+      factory: function() { return ['https:']; },
```
arobins: Is enforcing https on localhost worth the effort of setting up a certificate that works? My comment was more aimed at allowing things like https://storage.googleapis.com/web-api-confluence-data-cache|-dev/-staging|/whatever for doing potentially unsafe things without affecting production traffic aimed at .../web-api-confluence-data-cache/whatever.

mdittmer: My intention with these defaults was to choose "safe defaults"; i.e., it seems reasonable to trust encrypted data from `localhost` or `storage.googleapis.com`. When using localhost (as in some of my tests), I have been overriding the `safeProtocols` default to

PR #224, opened 2017-10-23
Simple web service DAOs: No syncing or versioning

mdittmer:
This removes all syncing/versioning logic from web service DAOs, replacing it with a JSON file-backed scheme. The new scheme will be safe to run in production under the following constraints:

1. We add custom health checks to ensure that JSON data have been loaded before instances receive traffic (to be in a follow-up PR);
2. The release process includes pushing new service code whenever we push new data; (then, once the new instances are ready, manually redirect traffic to the new instances with new data).


lib/json_dao_container.es6.js
```diff
@@ -24,23 +65,24 @@ foam.CLASS({
...
-      },
-    },
-    {
-      name: 'stubFactory',
+      class: 'FObjectProperty',
+      of: 'org.chromium.apis.web.DAOContainerConfig',
+      name: 'config',
       factory: function() {
-        return this.StubFactorySingleton.create();
+        return this.DAOContainerConfig.create({
```
arobins: I'm not sure that this container is providing much value over embedding the fields on the class directly.

main/serve.js
```diff
@@ -37,23 +55,20 @@ let server = pkg.Server.create({
...
-  gcloudAuthEmail: credentials.client_email,
-  gcloudAuthPrivateKey: credentials.private_key,
-  gcloudProjectId: credentials.project_id,
-  logger: logger,
-}) : pkg.DAOContainer.create({
-  logger: logger,
+const modeString = process.argv[2];
+const mode = pkg.JsonDAOContainerMode.VALUES.filter(function(value) {
+  return value.name === modeString;
+})[0] || pkg.JsonDAOContainerMode.LOCAL;
```
arobins: I think erroring out on bad input is better than defaulting to local.

lib/json_dao_container.es6.js
mdittmer: Removed the Config class.

main/serve.js
mdittmer: Done.

mdittmer:
@arobins, to get `serve.js` working correctly again, I applied some bug fixes. PTAL and LMK if you want any of these changes punted to a separate PR. They're not very big.

main/serve.js
```diff
@@ -32,28 +50,35 @@ foam.CLASS({
...
-});
-
 const logger = foam.log.ConsoleLogger.create();
-let credentials = null;
-try {
-  credentials = require('../.local/credentials.json');
-} catch (e) {
-  logger.warn(`No Cloud Datastore credentails found. Using read-only copy of
-                   local data snapshot`);
+const modeString = process.argv[2];
```
arobins: Position-based command lines are fairly brittle. I don't know if there's a good (or any) flag parsing libraries, and I wouldn't recommend writing one for this, so there may not be anything to be done here.

Either way, it can be in a later change.

PR #225, opened 2017-10-27
Faster ObjectGraph => JSON

mdittmer:
Run main ObjectGraph => JSON transformations in parallel. The new limiting factor is indexing records output by the transformation so they can be sorted before being stored.


lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -45,38 +54,116 @@ foam.CLASS({
...
+        });
+      },
+    },
+    {
+      class: 'FObjectProperty',
+      of: 'foam.box.Context',
+      name: 'apiContext_',
+      factory: function() {
+        let ctx = this.Context.create({
+        unsafe: false,
```
arobins: +2 indent here and next two lines

lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -45,38 +54,116 @@ foam.CLASS({
...
+        const numWorkers = this.numWorkers;
+        let workers = new Array(numWorkers);
+
+        for (let i = 0; i < numWorkers; i++) {
+          // Stub BoxRegistry interface over box addressed to new fork.
+          workers[i] = this.stubFactory.get(this.BoxRegistry).create({
+            delegate: this.ForkBox.create({
+              critical: true,
+              nodeParams:
+              // ['--inspect', '--debug-brk'],
```
arobins: Don't think you need the linebreak here

lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -45,38 +54,116 @@ foam.CLASS({
...
   ],
+
   methods: [
     {
       name: 'import',
       documentation: `Reads object graph files from objectGraphPath
           and extract web interfaces and import it to cloudstoreDAO
           using apiImporter.`,
       code: function() {
+        const ogFileRegExp = /window_[a-zA-Z0-9_.]+_[a-zA-Z0-9_.]+_[a-zA-Z0-9_.]+_[a-zA-Z0-9_.]+[.]json$/;
```
arobins: This makes me think that the collection of filenames could be its own method, or maybe an input to this object. The filename schema has to be right for the parsing to work, which is a bit ugly, but I think that can be addressed through documentation.

main/og_to_json.es6.js
```diff
@@ -0,0 +1,84 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+'use strict';
+
```
arobins: Document what this script does.

lib/web_catalog/object_graph_importer.es6.js
mdittmer: Done (next push).

lib/web_catalog/object_graph_importer.es6.js
mdittmer: This was for debugging. Done (next push).

lib/web_catalog/object_graph_importer.es6.js
mdittmer: I've done my best to document this where `objectGraphPath` is defined. I got halfway through refactoring this into accepting an array of paths before realizing that the file name format is being used to extract information, too. I think if this importer gets refactored in the future, that would be a good time to consider other ways to extract the information and make this interface more robust.

main/og_to_json.es6.js
mdittmer: Done.

mdittmer:
Comments addressed (and/or responded inline). @arobins PTAL.

PR #226, opened 2017-10-31
Faster metric computation

mdittmer:
Use a local-runner + (replicated) remote-service strategy to parallelize metric computation.

@arobins PTAL after #225 lands. (This currently has #225 changes merged in.)


lib/confluence/api_velocity.es6.js
```diff
@@ -31,90 +29,73 @@ foam.CLASS({
...
-        var promises = [this.computeForSingleRelease_(releases[0])];
-        for (var i = 0; i < releases.length - 1; i++) {
-          promises.push(this.computeForPair_(releases[i], releases[i + 1]));
-        }
-        return Promise.all(promises);
+      code: function(releases, date) {
+        const numReleases = releases.length;
+        foam.assert(numReleases === 1 || numReleases === 2,
+                    'ApiVelocity.compute(): Expected 1 or 2 releases');
+        const lastReleaseDate = releases[releases.length -1].releaseDate;
```
arobins: nit: format around -1

lib/confluence/api_velocity.es6.js
```diff
@@ -31,90 +29,73 @@ foam.CLASS({
...
-            release.releaseDate, release, [], []);
-        return release.interfaces.dao.select(this.COUNT())
-            .then(count => this.apiVelocityDAO.put(this.ApiVelocityData.create({
-              releaseDate: release.releaseDate,
-              browserName: release.browserName,
-              currRelease: release,
-              totalApis: count.value,
-            })));
+        this.info(`Computing ApiVelocity for single release ${release.id}`);
+        return this.recontextualize(release).interfaces.dao.select(this.COUNT())
```
arobins: recontextualize feels like a thing that should be done by the user of this object. If I'm following things correctly, this is being done because the service might have been instantiated with a different context than the code calling the service. Why isn't something in the service layer dealing with this? Putting it there means that this code can be used without the service wrapper in a way that doesn't copy every input object.

mdittmer: Done. Also fixed contextualization bug in `*_json_dao` that was causing some instances of "recontextualization needed".

mdittmer:
@jacksonic PTAL. @arobins took a first pass at this, but he's out for a few weeks.

PR #228, opened 2017-11-06
Server update

mdittmer:
Refactor server code to use newly rationalized FOAM2 HTTP server handlers. This change also pushes configuration out of *doesn't-show-up-in-test-coverage* `main/`.


scripts/serve.sh
```diff
@@ -34,13 +34,15 @@ function stop() {
...
+mkdir -p "${WD}/../.local"
+
 warn "STARTING WEBPACK"
 webpack --watch --progress --config "${WD}/../config/webpack.dev.js" &
 Webpack_PID=$!
 win "WEBPACK STARTED (PID=${Webpack_PID})"
 
 warn "STARTING WEB SERVER"
-node --max_old_space_size=4000 "${WD}/../main/serve.js" "LOCAL" &
+node --max_old_space_size=4000 "${WD}/../main/serve.js" "LOCAL" "DEV" &
```
jacksonic: Is the "DEV" intentional here for local running only? (also are the main readme's running instructions up to date?)

mdittmer: Yes, and no. This script is not invoked in deploying to production.

PR #230, opened 2017-11-06
Store HTTP JSON DAO base URL as data

mdittmer:


mdittmer:
Note: Current diffbase is `server-update`. Will change once #228 lands.

PR #231, opened 2017-11-06
Declarative indexing for OOP MDAOs

mdittmer:
Support `indices` definition that creates MDAO indexes, plumbed from specs in main front end server process to child processes hosting DAOs.


lib/indexed_dao.es6.js
```diff
@@ -0,0 +1,76 @@
...
+      in "of") and indexed according to "indices" property.`,
+
+  properties: [
+    {
+      class: 'Class',
+      name: 'of',
+    },
+    {
+      class: 'Array',
+      name: 'indices',
```
jacksonic: Since this is actually holding a list of properties to index (and not indexes themselves) I'd rename it to avoid confusion with the MDAO's actual root index.

mdittmer:
Renamed `indices` to `propertyIndexGroups`. @jacksonic PTAL.

PR #233, opened 2017-11-07
Define request cache

mdittmer:
Use a custom DAO for request-level caching: Do not store non-success responses.


mdittmer:
Yeah. I was considering making caches store  only `2xx` responses by default, but I figured I should keep the default strategy simple. Some kind of `EasyHandler` could have a flag for this common pattern (upstream).

PR #245, opened 2017-11-17
Prefer Windows release as basis for metric computations

mdittmer:
We were getting inaccurate metrics when, by chance, the OSX release of Firefox would be used to calculate some metrics. This ensures that a Windows release is used consistently as the "base release" for metrics computations (except, of course, when there is no Windows release: Safari/WebKit).

### DIFF
--- a/lib/confluence/metric_computer_runner.es6.js
+++ b/lib/confluence/metric_computer_runner.es6.js
@@ -86,10 +86,14 @@ foam.CLASS({
                 this.Release.BROWSER_NAME,
                 this.ArraySink.create()))
             .then((groups) => {
-          return groups.groupKeys.map((bName) => {
-            return groups.groups[bName].array[0];
-          });
-        });
+              return groups.groupKeys.map((bName) => {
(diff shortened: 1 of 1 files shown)


foolip:
This is already included in what's deployed to http://web-confluence.appspot.com, right?

lib/confluence/metric_computer_runner.es6.js
```diff
@@ -86,10 +86,14 @@ foam.CLASS({
                 this.Release.BROWSER_NAME,
                 this.ArraySink.create()))
             .then((groups) => {
-          return groups.groupKeys.map((bName) => {
-            return groups.groups[bName].array[0];
-          });
-        });
+              return groups.groupKeys.map((bName) => {
```
foolip: FWIW this had me wondering just a little bit if I might be inside some kind of sorting method, and if there's an `aName`. But I think this is the browser name. Expanding it wouldn't hurt.

mdittmer:
@foolip Re: deployed? Yes.

PR #246, opened 2017-11-18
StatsController for time-to-data-arrival and other things

mdittmer:
Initial metric is time-to-data-arrival on the catalog page. For now, the controller uses `window.location.hash` when it's instantiated to determine whether or not the statistic should apply. This is important because there is no navigation+reload when moving between front page, catalog page, and metrics page. The metric is only reported when the user first loads the page looking at a catalog (and non-empty catalog data eventually arrive).

### DIFF
--- a/lib/client/stats_controller.es6.js
+++ b/lib/client/stats_controller.es6.js
@@ -1,0 +1,92 @@
+// Copyright 2017 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+'use strict';
+
+foam.CLASS({
+  package: 'org.chromium.apis.web',
+  name: 'StatsController',
+
+  documentation: `Controller for reporting website statistics.`,
+
+  imports: [
+    'info',
+    'window?',
+  ],
+
+  topics: [
+    'displayMatrix',
+    'mergedDisplayMatrix',
+  ],
+
+  properties: [
+    {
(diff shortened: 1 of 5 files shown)


lib/client/stats_controller.es6.js
```diff
@@ -0,0 +1,92 @@
...
+      name: 'initialHash',
+      documentation: `URL hash part upon initialization, if applicable.`,
+    },
+  ],
+
+  methods: [
+    function init() {
+      this.SUPER();
+
+      // onDisplayMatrix dispatches 'displayMatrix' to meged listener:
```
arobins: typo: merged

lib/client/stats_controller.es6.js
```diff
@@ -0,0 +1,92 @@
...
+      this.displayMatrix.sub(this.onDisplayMatrix_);
+      // onDisplayMatrix_ dispatches 'mergedDisplayMatrix' to one-time listener:
+      // initialDisplayMatrix.
+      this.mergedDisplayMatrix.sub(foam.events.oneTime(
+          this.initialDisplayMatrix));
+
+      // Store initial location hash (if not injected). This signals some
+      // listeners as to whether their data are meaningful (i.e., whether the
+      // "right page" was loaded at time=0).
+      if ((!this.initialHash) && this.window && this.window.location) {
```
arobins: nit: remove brackets

PR #248, opened 2018-01-11
Add QQ release dates to version_history.json

foolip:
These were found starting at https://web.archive.org/web/20160801000000*/mb.qq.com

The release date is found by clicking Android, where it will say
"最新版本：8.0" + "更新时间：2017-11-22" or similar.

I also downloaded the APKs ("直接下载安装包") with the exception of
qqbrowser_7.5.0.3240_20820.apk, where the link no longer works.

7.2.1: https://web.archive.org/web/20170306123755/http://mb.qq.com/
7.3.1: https://web.archive.org/web/20170320120425/http://mb.qq.com/
7.5: https://web.archive.org/web/20170517120609/http://mb.qq.com:80/
7.5.1: https://web.archive.org/web/20170611032215/http://mb.qq.com/
7.6.1: https://web.archive.org/web/20170619151649/http://mb.qq.com:80/
7.8: https://web.archive.org/web/20170921113032/http://mb.qq.com/
8.0: https://web.archive.org/web/20171127152249/http://mb.qq.com/
8.1.3: https://web.archive.org/web/20180111102656/http://mb.qq.com/

There are undoubtedly gaps in the releases here, but unless there are
interesting changes between some of these versions, those gaps won't really
matter.


foolip:
> We do need to make sure that file names and UA info encoded in object graphs are consistent with the data we use for computing APIs.

Just file names, or really *and UA info* within?

mdittmer:
Unfortunately *UA info* within. The way this is structured right now is thoroughly broken if we cannot trust the UA string:

The `ObjectGraph` [lazily computes platform info from its copy of the UA string](https://github.com/mdittmer/object-graph-js/blob/b1e5e9fcd1c0768401fcfb08d6c2ca64251255ee/src/ObjectGraph.es6.js#L132). The `/save` endpoint ignores the `ObjectGraph`'s UA string [uses the client's actual UA string, to construct a filename for the data](https://github.com/mdittmer/web-apis/blob/master/main/serve.js#L174).

foolip:
Hmm, OK, some tweaking required then :(

PR #253, opened 2018-01-15
Browser/OS data from ObjectGraph JSON

mdittmer:
... fallback on old behaviour: Use filename. In combination with `mdittmer/web-apis` and `mdittmer/object-graph-js` changes, this allows data collectors to manually rewrite browser info (because some browsers do not contain precise enough info in their UA string), and for that info to propagate up into Confluence's data objects.

Dependent PRs:

https://github.com/mdittmer/web-apis/pull/55 + https://github.com/mdittmer/web-apis/pull/56
https://github.com/mdittmer/object-graph-js/pull/21


lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -134,31 +134,59 @@ foam.CLASS({
...
+                      ${releaseInfo.browser.name} ${releaseInfo.browser.version}
+                      ${releaseInfo.platform.name}
+                      ${releaseInfo.platform.version}`);
                 }));
           }
         }
         return Promise.all(promises);
       },
     },
+    function getReleaseInfo(filename) {
```
arobins: Should this be private?

lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -134,31 +134,59 @@ foam.CLASS({
           if (stat.isFile()) {
```
arobins: Does the pattern on 128 still make sense?

lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -134,31 +134,59 @@ foam.CLASS({
...
+                      ${releaseInfo.platform.name}
+                      ${releaseInfo.platform.version}`);
                 }));
           }
         }
         return Promise.all(promises);
       },
     },
+    function getReleaseInfo(filename) {
+      const json = require(`${this.objectGraphPath}/${filename}`);
```
foolip: This is neat, I hadn't seen this pattern before. https://goenning.net/2016/04/14/stop-reading-json-files-with-require/ was the 3rd result when searching "nodejs require json" to figure out if this is a common thing, but I don't believe the caching should be a problem here, nor the sync I/O.

lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -134,31 +134,59 @@ foam.CLASS({
...
+      if (json.environment && json.environment.browser &&
+          json.environment.browser.name && json.environment.browser.version &&
+          json.environment.platform && json.environment.platform.name &&
+          json.environment.platform.version) {
+        return this.getReleaseInfoFromJSON_(json);
```
foolip: The if statement has such exhaustive knowledge of what `getReleaseInfoFromJSON_` needs that I think just inlining `json.environment` would be more clear. Or move the conditions into `getReleaseInfoFromJSON_` and throw an exception or return null if it's not met, and handle that here.

lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -134,31 +134,59 @@ foam.CLASS({
...
+        return this.getReleaseInfoFromFilename_(filename);
+      }
+    },
+    function getReleaseInfoFromJSON_(json) {
+      return json.environment;
+    },
+    function getReleaseInfoFromFilename_(filename) {
+      let ret = {browser: {}, platform: {}};
+
+      // Drop slice(0, -5): ".json".
```
foolip: `foam.assert(filename.endsWith('.json'))` perhaps?

lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -134,31 +134,59 @@ foam.CLASS({
...
+      const releaseInfo = filename.slice(0, -5).split('_');
+      ret.browser.name = releaseInfo[1];
```
foolip: If so, while asserting things, also assert `releaseInfo.length == 5`, which might also calm the reader that `releaseInfo[0]` is knowingly ignored. Or, if you like it, `[_, ret.browser.name, ret.browser.version, ret.platform.name, ret.platform.version] = releaseInfo;`.

lib/web_catalog/object_graph_importer.es6.js
mdittmer: No. Good point.

lib/web_catalog/object_graph_importer.es6.js
mdittmer: No, in the sense that it should be a "testable unit". Yes in the sense that nothing else calls it.

lib/web_catalog/object_graph_importer.es6.js
mdittmer: I'll fix it anyway.

lib/web_catalog/object_graph_importer.es6.js
mdittmer: Good point.

lib/web_catalog/object_graph_importer.es6.js
mdittmer: Indeed.

lib/web_catalog/object_graph_importer.es6.js
mdittmer: Done.

mdittmer:
**ping** @arobins

test/node/object_graph_importer-test.es6.js
```diff
@@ -0,0 +1,47 @@
...
+    expect(releaseInfo.platform.name).toBe('OSX');
+    expect(releaseInfo.platform.version).toBe('10.11');
+  });
+
+  it('should use data from "environment" when it is available', () => {
+    const ogi = ObjectGraphImporter.create({objectGraphPath});
+
+    // Test file contains Chrome63/Linux64 data with browser and platform names
+    // overridden as "Test".
+    const releaseInfo = ogi.getReleaseInfo('window_IgnoreThisNameEntirely.json');
```
arobins: The test file for this and the next test probably don't need to be full data dumps.

lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -134,34 +134,65 @@ foam.CLASS({
...
+                      ${releaseInfo.platform.version}`);
                 }));
           }
         }
         return Promise.all(promises);
       },
     },
+    function getReleaseInfo(filename) {
+      const json = JSON.parse(fs.readFileSync(
+          `${this.objectGraphPath}/${filename}`));
```
arobins: This should be path.join'd instead of concatenated. I'd also try to pass in the full path instead of the name to minimize the number of places where you are building the path.

lib/web_catalog/object_graph_importer.es6.js
```diff
@@ -134,34 +134,65 @@ foam.CLASS({
...
+                      ${releaseInfo.platform.name}
+                      ${releaseInfo.platform.version}`);
                 }));
           }
         }
         return Promise.all(promises);
       },
     },
+    function getReleaseInfo(filename) {
+      const json = JSON.parse(fs.readFileSync(
```
arobins: This means that each imported file will be loaded into memory and parsed twice, right?

lib/web_catalog/object_graph_importer.es6.js
arobins: I would have gone further and said that the name format doesn't matter at all.

lib/web_catalog/object_graph_importer.es6.js
mdittmer: Sad, but true. We temporarily extract the file to get release info, then throw it away. Later, worker processes extract it for API extraction.

arobins: Would it be unreasonable for the API extractor to also return the release info? It already has the information (file content or file name).

lib/web_catalog/object_graph_importer.es6.js
mdittmer: The downside of that approach is that we then need to extract the filename for (or inside) `getReleaseInfoFromFilename_()`.

lib/web_catalog/object_graph_importer.es6.js
mdittmer: The trouble with that is that if anything else happens to live in that directory, we attempt to interpret it as valid JSON. I've run into that problem before.

lib/web_catalog/object_graph_importer.es6.js
arobins: Calling path.basename(filepath) seems less prone to breakage than building the path in two places.

test/node/object_graph_importer-test.es6.js
mdittmer: Probably not, but the test run time cost does not appear to be high and I'd prefer to run the tests against "real data". WDYT?

lib/web_catalog/object_graph_importer.es6.js
mdittmer: That seems reasonable. Just a bigger change because right now its return value is an API listing (with no metadata).

mdittmer:
**ping** @arobins

PR #254, opened 2018-01-16
Initial GridDAO implementation

mdittmer:
This will be used to represent Release<-->API data as a big table of bools which can be skipped/limited and "projected" (to filter out releases not currently in view). It is hoped that this representation can retire API Matrix, simplify client code, and decrease time-to-data.


lib/grid_dao.es6.js
```diff
@@ -0,0 +1,51 @@
...
+      class: 'Array',
+      documentation: 'Column metadata items parallel to GridRow.data.',
+      name: 'cols',
+      final: true,
+      required: true,
+    },
+  ],
+
+  methods: [
+    function colIndexOf(col) {
```
arobins: Documentation

PR #255, opened 2018-01-16
Test & implement: Projection over data columns in GridDAO sink

mdittmer:
This change provides a convenient API for sending `GridDAO` queries to go to a `Sink` that extracts particular array elements from the `GridRow.data`.


lib/grid_dao.es6.js
```diff
@@ -18,18 +18,67 @@ foam.CLASS({
   ],
 });
 
+
+foam.CLASS({
+  package: 'org.chromium.mlang.sink',
+  name: 'Projection',
+  extends: 'foam.mlang.ExprProperty',
+
+  documentation: `Projection over particular data columns. Primarliy used to
```
arobins: nit: typo in primarily

lib/grid_dao.es6.js
```diff
@@ -18,18 +18,67 @@ foam.CLASS({
...
+  properties: [
+    {
+      class: 'Array',
+      of: 'Int',
+      name: 'cols',
+    },
+    {
+      name: 'f',
+      value: function f(gridRow) {
+        let row = gridRow.clone();
```
arobins: It might be better to only copy the elements that are needed, and then construct a row that wraps the new array.

test/any/grid_dao-test.es6.js
```diff
@@ -81,4 +81,102 @@ describe('GridDAO', () => {
...
+      expect(array[1].id).toBe(1);
+      expect(array[2].id).toBe(2);
+      expect(array[3].id).toBe(3);
+      expect(array[0].data).toEqual([0]);
+      expect(array[1].data).toEqual([1]);
+      expect(array[2].data).toEqual([2]);
+      expect(array[3].data).toEqual([3]);
+      done();
+    });
+  });
```
arobins: Consider adding a test for dropped columns.

lib/grid_dao.es6.js
mdittmer: Interesting point. The only trade-off I see there is that if `GridRow` is later changed to have more properties, then this code needs to be updated to stay in sync.

test/any/grid_dao-test.es6.js
mdittmer: Do you mean include a test that leaves out some columns (rather than re-ordering 100% them)? Or do you mean something else?

arobins: Yes, a test that leaves out some columns.

lib/grid_dao.es6.js
arobins: That is the downside. If you have enough data in the grid, there's probably a measurable cost to copying the whole thing and then GCing the discarded portion of it.

lib/grid_dao.es6.js
mdittmer: Done.

lib/grid_dao.es6.js
mdittmer: Created a new entity copying only ID (then set data when it's ready). In case of a modeled ID, I attempt to clone the ID (and otherwise use assignment).

test/any/grid_dao-test.es6.js
mdittmer: Done.

PR #266, opened 2018-03-16
Updates to fix data collection

mdittmer:
- Make NodeJS `require()` properties transient
- Update `package.json` and lock in version with `package-lock.json`


mdittmer:
Yes. It provides a version lock for dependencies, on top of the more relaxed version constraints in `package.json`.

PR #267, opened 2018-03-16
New CLI entry point: Relational JSON => GridRow JSON transform

mdittmer:


main/relational_to_grid.es6.js
```diff
@@ -0,0 +1,115 @@
...
+        BaseConfluenceDataURL = absolute https: or file: URL to directory
+                                where Confluence data are stored with NO
+                                trailing slash.`;
+if (process.argv.length !== 3) {
+  console.error(USAGE);
+  process.exit(1);
+}
+
+const dataUrl = url.parse(process.argv[2]);
+if (dataUrl.protocol !== 'file:' && dataUrl.protocol !== 'https:') {
```
arobins: Because a spec that says 'append character : to every string returned by this method' is a good spec, right?

Or, just another wat from JS-land.

main/relational_to_grid.es6.js
```diff
@@ -0,0 +1,115 @@
...
+    return new Promise((resolve, reject) => {
+      require('fs').writeFile(
+          `${__dirname}/../data/json/${cls.id}.json`,
+          outputter.stringify(arraySink.array, cls),
+          error => {
+            if (error) {
+              logger.error(`Error storing ${cls.id}`, error);
+              reject(error);
+            } else {
+              logger.info(`Stored ${cls.id}`, error);
```
arobins: error probably isn't a useful thing to log here because it's always something false-like

main/relational_to_grid.es6.js
```diff
@@ -0,0 +1,115 @@
...
+require('../lib/web_apis/web_interface.es6.js');
+const pkg = org.chromium.apis.web;
+
+const logger = foam.log.ConsoleLogger.create();
+
+const USAGE = `USAGE:
+
+    node /path/to/import.es6.js BaseConfluenceDataURL
+
+        BaseConfluenceDataURL = absolute https: or file: URL to directory
```
arobins: This script could use some documentation about what it does.

main/relational_to_grid.es6.js
```diff
@@ -0,0 +1,115 @@
...
+  const joins = arraySinks[2].array;
+
+  const outputter = foam.json.Outputter.create({
+    pretty: false,
+    formatDatesAsNumbers: true,
+    outputDefaultValues: false,
+    useShortNames: false,
+    strict: true,
+  });
+  function store(arraySink) {
```
arobins: Maybe this is just a promise-ism/JSism, but this looks like it could have been defined outside of the promise, and made following along with the logic of the promise a little easier.

main/relational_to_grid.es6.js
```diff
@@ -0,0 +1,115 @@
...
+    const cls = arraySink.of || arraySink.array[0] ? arraySink.array[0].cls_ :
+        foam.core.FObject;
+    logger.info(`Storing ${cls.id}`);
+    return new Promise((resolve, reject) => {
+      require('fs').writeFile(
```
arobins: I think you want to require this at the top like the other modules. Failing early is better.

main/relational_to_grid.es6.js
mdittmer: I'm not sure I understand the question. URLs could have other schemes like `http:` or `chrome:`. We don't want to attempt to use them in this script.

main/relational_to_grid.es6.js
mdittmer: Done (push pending).

main/relational_to_grid.es6.js
mdittmer: I think I follow what you mean. I moved `outputter` and `store` to the global scope/context.

main/relational_to_grid.es6.js
mdittmer: Done.

main/relational_to_grid.es6.js
mdittmer: Done (deleted `, error`)

main/relational_to_grid.es6.js
arobins: Sorry, just commenting on silly things that JS does. The spec for this field says that all implementations must include a (useless) colon to the end of the scheme string.

PR #268, opened 2018-03-17
Fix tests: Update FOAM and multi-part-id

mdittmer:


mdittmer:
Very strange that coverage decreases. I guess different/less Confluence code is exercised on `find()` now?

PR #269, opened 2018-03-23
Port custom release-bound property and abstract compat data class…

mdittmer:
… from `mdittmer/mdn-confluence`

[Property class](https://github.com/mdittmer/mdn-confluence/blob/39dad352ee0316cbfb77f963e372d9295972644a/public/lib/org/mozilla/mdn/GridProperty.es6.js)
[Abstract basis for future generated class](https://github.com/mdittmer/mdn-confluence/blob/39dad352ee0316cbfb77f963e372d9295972644a/public/lib/org/mozilla/mdn/ConfluenceClassGenerator.es6.js#L24)


mdittmer:
//CC @lukebjerring

PR #270, opened 2018-03-26
Port compat data model generator from mdittmer/mdn-confulence

mdittmer:
Generates a custom class for compat data over a specific set of releases.

### DIFF
--- a/lib/web_apis/api_compat_data.es6.js
+++ b/lib/web_apis/api_compat_data.es6.js
@@ -136,7 +136,6 @@ foam.CLASS({
       documentation: `Generate the "spec" that would be passed to foam.CLASS()
           to declare a class of given [pkg].[name], based on given releases.`,
       code: function(pkg, name, releases) {
-        const generator = this;
         let spec = {
           class: 'Model',
           package: pkg,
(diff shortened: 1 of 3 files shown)


lib/web_apis/api_compat_data.es6.js
```diff
@@ -120,3 +120,96 @@ foam.CLASS({
...
+
+  axioms: [foam.pattern.Singleton.create()],
+
+  methods: [
+    {
+      name: 'generateSpec',
+      documentation: `Generate the "spec" that would be passed to foam.CLASS()
+          to declare a class of given [pkg].[name], based on given releases.`,
+      code: function(pkg, name, releases) {
+        const generator = this;
```
arobins: Unused

lib/web_apis/api_compat_data.es6.js
```diff
@@ -120,3 +120,96 @@ foam.CLASS({
...
+          // Declare class from spec as usual.
+          const spec = specOrModel;
+          foam.CLASS(spec);
+          return foam.lookup(`${spec.package}.${spec.name}`);
+        }
+      },
+    },
+    {
+      name: 'propertyNameFromRelease',
+      documentation: `Produce a consistent camelCase property name from a
```
arobins: nit: camelcase names don't typically contain underscores.

lib/web_apis/api_compat_data.es6.js
```diff
@@ -120,3 +120,96 @@ foam.CLASS({
...
+          release object. The property is used to store compat data for the
+          release.`,
+      code: function (release) {
+        return release.browserName.charAt(0).toLowerCase() +
+            release.browserName.substr(1).replace(/[^a-z]/ig, '_') +
```
arobins: Might be easier to do the concats first, then sanitize the whole string. I don't think stripping numbers out of browser and OS name is worth the extra regexp passes.

test/any/web_apis/api_compat_data-test.es6.js
```diff
@@ -56,3 +56,94 @@ describe('AbstractApiCompatData', () => {
...
+      osVersion: '7.8.9',
+    });
+    const Cls = generator.generateClass(
+        generator.generateSpec('org.chromium.apis.web', 'CompatClass', [
+          release0,
+          release1,
+        ]));
+    const props = Cls.getAxiomsByClass(CompatProperty);
+    expect(props.length).toBe(2);
+    expect(foam.util.equals(props[0].release, release0)).toBe(true);
```
arobins: Add verification of property names

mdittmer: My thinking here was that the naming convention is an implementation detail. The thing we care to test is the association with the releases. I.e., suppose some weird use of characters in browser versions causes us to change the naming convention; not sure that should require an update of this test.

I suppose I can see it the other way though: something that requires a change in the naming convention probably surfaces as a bug, and we should lock in the fix with a (modified) test.

arobins: It depends on how this is going to be used. Unintentionally changing things that other code depends on would be bad, which a test for the name generator would be more likely to catch.

mdittmer: I added some more expectations for this.

PR #271, opened 2018-03-26
Script and singleton for relational => compat code and data generation

mdittmer:


mdittmer:
FYI: @arobins and @lukebjerring no tests for this yet, but I intend to add some to this PR. I'll push tests for this PR, then it'll be fully ready for review.

mdittmer:
Tests pushed. @arobins PTAL.

lib/web_apis/relational_to_compat.es6.js
```diff
@@ -0,0 +1,84 @@
...
+      name: 'convert',
+      code: function(releases, apis, joins) {
+        this.info('Generating data model for ${releases.length} releases');
+        const generator = this.classGenerator;
+        const CompatData = generator.generateClass(generator.generateSpec(
+            'org.chromium.apis.web.generated', 'CompatData', releases));
+        this.info('Data model generated');
+
+        const compatProps = CompatData.getAxiomsByClass(this.CompatProperty);
+        const setCompatBit = (data, releaseId) => {
```
arobins: Maybe move this closer to usage, or just inline it because it's only used in one place.

lib/web_apis/relational_to_compat.es6.js
```diff
@@ -0,0 +1,84 @@
...
+          const prop = compatProps.filter(p => p.release.id === releaseId)[0];
```
arobins: Is this fast enough to do once per join entry because the release list is expected to be small?

main/relational_to_compat.es6.js
```diff
@@ -0,0 +1,144 @@
...
+[class package path].json in [repository root]/data, with class model files in
+class:[class package path].json.
+
+USAGE:
+
+    node /path/to/relationship_to_compat.es6.js BaseConfluenceDataURL
+
+        BaseConfluenceDataURL = absolute https: or file: URL to directory
+                                where Confluence data are stored with NO
+                                trailing slash.`;
```
arobins: Should the tool verify that there's no trailing slash on the input? Alternatively, the reads will (probably) fail quickly if you've provided a bad path, so maybe this note isn't important?

lib/web_apis/relational_to_compat.es6.js
mdittmer: Done

lib/web_apis/relational_to_compat.es6.js
mdittmer: Good point. I'll use object-as-map.

main/relational_to_compat.es6.js
mdittmer: Maybe there should be verification. My thinking was that if reads fail with a trailing slash, but you fail to notice the `//` showing up in logging of URLs, at least the usage string will provide a hint as to why it's failing.

main/relational_to_compat.es6.js
```diff
@@ -0,0 +1,144 @@
...
+if (process.argv.length !== 3) {
+  console.error(USAGE);
+  process.exit(1);
+}
+
+const dataUrl = url.parse(process.argv[2]);
+if (dataUrl.protocol !== 'file:' && dataUrl.protocol !== 'https:') {
```
lukebjerring: nit: `!['https:', 'file:'].includes(dataUrl.protocol)`

PR #273, opened 2018-03-28
Port query parser/interpreters from mdittmer/mdn-confluence

mdittmer:
This code is reasonably well tested, but one test is currently failing. We need to decide whether or not the partial parse is "good enough" or whether we have to muck with the grammar some more.

### DIFF
--- a/lib/parse/parser_interpreter.es6.js
+++ b/lib/parse/parser_interpreter.es6.js
@@ -135,11 +135,9 @@ foam.CLASS({
       }
       const columns = this.getMatchingColumns(key);
       if (columns.length === 0) return ret(null);
-      // Starts with 't' or non-zero digit => true;
-      // Starts with 'f' or '0' => false;
-      // Otherwise pass to delegate.
-      const boolValue = /^[t1-9]/i.test(value) ? true :
-              /^(f.*|0)$/i.test(value) ? false : null;
+      const lowerValue = value.toLowerCase();
+      const boolValue = lowerValue === 'true' ? true :
+            (lowerValue === 'false' ? false : null);
       if (boolValue === null) return ret(null);
 
       if (columns.length === 1) {
--- a/lib/parse/parser_interpreter.es6.js
+++ b/lib/parse/parser_interpreter.es6.js
@@ -7,205 +7,15 @@ require('../grid_dao.es6.js');
 require('../web_apis/api_compat_data.es6.js');
 require('./expressions.es6.js');
 
-foam.INTERFACE({
-  name: 'QueryInterpreter',
-  package: 'org.chromium.parse',
-
-  documentation: 'Interface for applying semantics to parsed query fragments.',
-
-  properties: [
-    {
-      class: 'FObjectProperty',
-      of: 'org.chromium.parse.QueryInterpreter',
-      documentation: 'Support proxying/delegating by default.',
-      name: 'delegate',
-      value: null,
-    },
-  ],
-
-  methods: [
-    {
-      name: 'interpretKeyValue',
-      returns: 'foam.mlang.predicate.AbstractPredicate',
-      args: [
-        {
-          typeName: 'String',
-          documentation: 'The string key in a "key:value" query fragment.',
-          name: 'key',
-        },
-        {
-          typeName: 'String',
-          documentation: 'The string value in a "key:value" query fragment.',
-          name: 'value',
-        },
-      ],
-    },
-    {
-      name: 'interpretKeyword',
-      returns: 'foam.mlang.predicate.AbstractPredicate',
-      args: [
-        {
-          typeName: 'String',
-          documentation: 'The keyword in a keyword query fragment.',
-          name: 'kw',
-        },
-      ],
-    },
-  ],
-});
-
-foam.CLASS({
-  name: 'DefaultQueryInterpreter',
-  package: 'org.chromium.parse',
-  implements: [
-    'foam.mlang.Expressions',
-    'org.chromium.parse.QueryInterpreter',
-  ],
-
-  documentation: `Basic interpreter that matches key/values to properties and
-      interprets keywords in the natural way.`,
-
-  requires: ['foam.core.Property'],
-
-  methods: [
-    function interpretKeyValue(key, value, parser) {
-      const property = parser.of.getAxiomByName(key);
-      if (!property || !this.Property.isInstance(property)) {
-        return this.delegate ?
-            this.delegate.interpretKeyValue(key, value, parser) :
-            this.interpretKeyword(`${key}:${value}`, parser);
-      }
-      return this.EQ(property, value);
-    },
-    function interpretKeyword(kw, parser) {
-      return this.KEYWORD(kw);
-    },
-  ],
-});
-
-foam.CLASS({
-  name: 'ReleaseApiQueryInterpreter',
-  package: 'org.chromium.parse',
-  implements: [
-    'foam.mlang.Expressions',
-    'org.chromium.mlang.GridExpressions',
-    'org.chromium.mlang.ParseExpressions',
-    'org.chromium.parse.QueryInterpreter',
-  ],
-
-  documentation: `Interpreter for handling release key matching and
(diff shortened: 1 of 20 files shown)


lib/parse/parser_interpreter.es6.js
```diff
@@ -0,0 +1,299 @@
...
+    {
+      name: 'queryGrammar',
+      language: 'foam.parse.Parsers',
+      symbols: function() {
+        return {
+          START: sym('query'),
+          query: seq1(1, sym('ws'), sym('or'), sym('ws')),
+
+          or: plus(sym('and'), sym('orDelim')),
+          orDelim: alt(seq(sym('ws'), literalIC('OR'), sym('ws')),
```
arobins: seq(ws, alt(OR, |), ws) would be a slightly easier to follow way of writing this. Or just drop the | version, because and doesn't support &.

lib/parse/parser_interpreter.es6.js
```diff
@@ -0,0 +1,299 @@
...
+              this.delegate.interpretKeyValue(key, value, parser) : v) :
+          (v => v === null ?
+              this.interpretKeyword(`${key}:${value}`, parser) : v);
+
+      if (key === 'count') {
+        return ret(this.interpretCount(value, parser));
+      }
+      const columns = this.getMatchingColumns(key);
+      if (columns.length === 0) return ret(null);
+      // Starts with 't' or non-zero digit => true;
```
arobins: This feels unnecessarily complicated to me, but I have no context for why it's this way.

lib/parse/parser_interpreter.es6.js
```diff
@@ -0,0 +1,299 @@
...
+      // When delegate is set and interpretation is null, pass to delegate for
+      // interpretation.
+      const ret = this.delegate ?
+          (v => v === null ?
+              this.delegate.interpretKeyValue(key, value, parser) : v) :
+          (v => v === null ?
+              this.interpretKeyword(`${key}:${value}`, parser) : v);
+
+      if (key === 'count') {
+        return ret(this.interpretCount(value, parser));
```
arobins: You call the delegate twice here. Was that intentional?

lib/parse/parser_interpreter.es6.js
```diff
@@ -0,0 +1,299 @@
...
+        return this.DefaultQueryInterpreter.create();
+      },
+    },
+  ],
+
+  methods: [
+    function interpretKeyValue(key, value, parser) {
+      // When delegate is set and interpretation is null, pass to delegate for
+      // interpretation.
+      const ret = this.delegate ?
```
arobins: Swapping the order of the checks here and below would make it easier to read.

lib/parse/parser_interpreter.es6.js
```diff
@@ -0,0 +1,299 @@
...
+          keyValue: seq(sym('key'), ':', sym('value')),
+          key: str(plus(alt(range('a', 'z'), range('A', 'Z'),
+                            range('0', '9'), '.', '_'))),
+          value: str(plus(alt(range('a', 'z'), range('A', 'Z'),
+                              range('0', '9'), '.', '_', '#', ','))),
+
+          keyword: str(plus(alt(range('a', 'z'), range('A', 'Z'),
+                                range('0', '9'), '.', '_', '#'))),
+
+          ws: repeat0(chars(' \t\r\n')),
```
arobins: Would the grammar be any less complicated if you collapsed whitespace before feeding it to the parser? Since you don't have to deal with strings in your query language, collapsing them to a single space should be easy.

lib/parse/parser_interpreter.es6.js
```diff
@@ -0,0 +1,299 @@
...
+  documentation: `Interpreter for handling release key matching and
+      count-of-selected-releases. E.g.,
+
+      count:3 => "Exactly 3 selected releases support API"
+
+      fir59win10:0 => [prefix match alphabetical and numeric groups]
+                   => "Not shipped (last 0 is false) in Firefox 59 Windows 10"
+
+      chr60:t => [prefix match alphabetical and numeric groups]
+              => "Shipped (t for false) in Chrome 60 on all known platforms"`,
```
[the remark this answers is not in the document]
mdittmer: Yeah. t does not stand for false. :man_facepalming:

lib/parse/parser_interpreter.es6.js
mdittmer: Not really, just defensive programming. We will only invoke the delegate twice if both `this` and `delegate` produce `null` (causing fallback in `interpretCount` and in `ret`). It's defensive because if `interpretCount` is changed to no longer fall back on the delegate, it'll still happen at least once in this code path.

lib/parse/parser_interpreter.es6.js
mdittmer: This is just my wild guess as to how programmers might try to use a Boolean value label. `t`, `true`, `TRUE`, `truthy`, and `1` (as opposed to `0`) seemed like reasonable "guesses at how to construct a valid query". I'm open to feedback on why this might be confusing or dangerous.

lib/parse/parser_interpreter.es6.js
```diff
@@ -0,0 +1,299 @@
...
+
+  methods: [
+    function parseString(str, opt_name) {
+      return this.queryGrammar.parseString(str, opt_name);
+    },
+  ],
+
+  grammars: [
+    {
+      name: 'queryGrammar',
```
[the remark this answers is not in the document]
mdittmer: Machines and humans. We construct URLs that constitute queries for "APIs associated with [point] on [Confluence graph]". Humans will also be able to enter structured queries in the search box above the catalog table.

lib/parse/parser_interpreter.es6.js
mdittmer: Hmm... `'&'` was supposed to be supported. I don't remember why/whether there was a good reason it got dropped. Anyway, I added it back and switched to a more readable form here.

lib/parse/parser_interpreter.es6.js
mdittmer: I'm not convinced it would be substantially simpler. We would essentially replace `sym('ws')` with `optional(' ')`.

lib/parse/parser_interpreter.es6.js
arobins: They are already learning a new query language. Telling them what you want them to enter seems like a better approach than letting them put in random garbage with little to no guidance on what that garbage is going to do.

lib/parse/parser_interpreter.es6.js
```diff
@@ -0,0 +1,312 @@
...
+  methods: [
+    function parseString(str, opt_name) {
+      // Parse string; on complete parse, return result.
+      const result = this.queryGrammar.getParse(str, opt_name);
+      if (result && result.pos === str.length) return result.value;
+
+      // Store partial parse, and extract keywords from remaining string.
+      const partialParse = result && result.value;
+      const keywords = str.substr(partialParse ? result.pos : 0)
+            .replace(/[ \t\r\n]+/g, ' ').split(' ').filter(kw => !!kw);
```
arobins: From offline: filter could be replaced with trim to make it more explicit that the goal is to remove empty strings, since those empty strings can only be from leading and trailing whitespace after the collapsing done by the replace call.

lib/parse/parser_interpreter.es6.js
```diff
@@ -0,0 +1,312 @@
...
+      const keywordArgs = keywords.map(kw => this.KEYWORD(kw));
+
+      // Edge case: no partial parse or keywords? return TRUE.
+      if (!partialParse && keywords.length === 0) return this.TRUE;
+
+      // Return conjunction of partial parse and keywords.
+      const partialParseArgs =
+            foam.mlang.predicate.And.isInstance(partialParse) ?
+            partialParse.args : partialParse ? [partialParse] : [];
```
arobins: I'd do the null check first, then the instanceof check. Maybe add some brackets around the second ternary to make reading it easier.

PR #275, opened 2018-04-15
Port DAO controller from mdn-confluence

mdittmer:
The underlying views are tested using a manual UI test.

### DIFF
--- a/lib/u2/DAOControllerView.es6.js
+++ b/lib/u2/DAOControllerView.es6.js
@@ -13,6 +13,11 @@ foam.CLASS({
   extends: 'foam.u2.View',
   implements: ['foam.mlang.Expressions'],
 
+  documentation: `Default view for a DAOController. Renders a search box,
+      columns selector, DAO table view and a "n of m" record count: n = number
+      of records according to search filter; m = total number of records in
+      underlying DAO.`,
+
   requires: [
     'foam.u2.md.OverlayDropdown',
     'org.chromium.apis.web.ActionView',
@@ -111,12 +116,14 @@ foam.CLASS({
       class: 'FObjectProperty',
       of: 'org.chromium.apis.web.DAOController',
       name: 'data',
+      documentation: 'The DAOController being rendered by this view.',
       expression: function(importedData) { return importedData; },
       required: true,
     },
     {
       class: 'Class',
       name: 'of',
+      documentation: 'The class of objects stored in the underlying DAO.',
       // TODO(markdittmer): This should work as an expression, but doesn't.
       factory: function() { return this.data.data.of; },
     },
@@ -124,6 +131,8 @@ foam.CLASS({
       class: 'FObjectArray',
       of: 'foam.core.Property',
       name: 'selectable',
+      documentation: `The properties that the user can select for rendering in 
+          the DAO table view.`,
       factory: function() {
         const imported = this.importedSelectableColumns;
         if (Array.isArray(imported) && imported.length > 0 &&
@@ -144,6 +153,8 @@ foam.CLASS({
       class: 'FObjectArray',
       of: 'foam.core.Property',
       name: 'selected',
+      documentation: `The properties currently selected for rendering in the DAO
+          table view.`,
       factory: function() {
         return this.importedSelectedColumns ||
             Array.from(this.selectable);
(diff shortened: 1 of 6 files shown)


lib/u2/DAOControllerView.es6.js
```diff
@@ -0,0 +1,278 @@
...
+      this.onFilteredDAOChange();
+    },
+    function initE() {
+      let daoView = this.data.FILTERED_DAO.toE(null, this);
+      daoView.addClass(this.myClass('DAOView'));
+      let searchView = foam.u2.ViewSpec.createView(
+          this.searchSpec, {}, this, this.__subSubContext__)
+          .addClass(this.myClass('search'));
+      this.addClass(this.myClass())
+          .start('div').addClass(this.myClass('controls'))
```
arobins: There's something to be said about the readability of a 17 line statement. I'd probably start by breaking it up according to the nesting level in the resulting HTML, because trying to match start and end tags in this is more challenging than it should be.

lib/u2/ToggleListElementsView.es6.js
```diff
@@ -0,0 +1,234 @@
...
+          }
+          this.selected = Array.from(ed);
+        },
+        function removeElement() {
+          const elem = this.selectable[this.selectableIdx];
+          const ed = this.selected;
+          for (let i = 0; i < ed.length; i++) {
+            if (ed[i].name === elem.name) {
+              ed.splice(i, 1);
+              i--;
```
arobins: This method should only be able to remove one element, so you could break here, right?

And if so, could you not use find to get the index, and then feed that to splice?

lib/u2/ToggleListElementsView.es6.js
```diff
@@ -0,0 +1,234 @@
...
+                            label`);
+          }
+        }
+      },
+    },
+    {
+      class: 'Array',
+      name: 'selected',
+      postSet: function(_, nu) {
+        for (const elem of this.selectable) {
```
arobins: I think selectable should be selected

lib/u2/ToggleListElementsView.es6.js
```diff
@@ -0,0 +1,234 @@
...
+        function getElement() {
+          return this.selectable[this.selectableIdx];
+        },
+        function addElement() {
+          const elem = this.selectable[this.selectableIdx];
+          const able = this.selectable;
+          const ed = this.selected;
+          let iAble = 0;
+          let iEd = 0;
+          while (iAble <= this.selectableIdx) {
```
arobins: An easier to follow implementation could be to store the selected state of each element in the element, and then compute the selected array by filtering selectable on it for each update. That also prevents having duplicates added to selected, which this doesn't appear to guard against.

lib/u2/DAOControllerView.es6.js
```diff
@@ -0,0 +1,278 @@
...
+                .add(data.cls_.getAxiomsByClass(foam.core.Action));
+          }, this.data$))
+          .start('span').addClass(this.myClass('columns-container'))
+          .add(this.columnsIconE)
+          .add(this.columnSelectionE)
+          .end()
+          .end()
+          .add(daoView)
+          .start('div').addClass(this.myClass('count'))
+          .add(this.count_$)
```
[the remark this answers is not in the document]
mdittmer: Yeah. I'm open to suggestions. The intention here is "N match current filter, of M total". Maybe add the [MD filter icon](https://storage.googleapis.com/material-icons/external-assets/v4/icons/svg/ic_filter_list_black_24px.svg) beside?

lib/u2/DAOControllerView.es6.js
mdittmer: Indentation done.

lib/u2/ToggleListElementsView.es6.js
mdittmer: I guess I was going for remove all behaviour, but then the method name betrays the implementation. `removeElement`=remove first matching element found seems like a reasonable implementation.

mdittmer: Done.

lib/u2/ToggleListElementsView.es6.js
mdittmer: Yep. Done.

lib/u2/ToggleListElementsView.es6.js
mdittmer: This implementation was intended to be non-intrusive to the elements of `selected` and `selectable`. I'll have to think about the duplicate guard issue, though.

mdittmer: Added multiple-addElement() guard and more algorithm documentation. @arobins PTAL.

PR #277, opened 2018-04-28
Test and implement class generator for compat data

mdittmer:


main/ui_test.es6.js
```diff
@@ -81,130 +81,129 @@ global.uiTestEnvPromise = (async function() {
   const data = foam.dao.MDAO.create({
     of: Cls,
   });
-  await Promise.all(array.map(item => data.put(item)));
+  return Promise.all(array.map(item => data.put(item))).then(() => {
```
arobins: Am I correct in assuming that the changes below are entirely whitespace?

mdittmer: That's correct.

PR #281, opened 2018-05-03
 Serializable DAO for web worker

mdittmer:
This is part of the new data layer strategy. The current `api_service` will be reduced to a collection of DAOs that each initially delegate to a `RestDAO` speaking directly to the network. Each will also register a `WorkerDAO`in a worker. Once the `WorkerDAO` cache is full, the service's proxy will point to it instead.

This PR just includes the serializable `WorkerDAO` piece (not the higher level proxy-to-network-until-cache-ready piece that will run on the page's main thread).


lib/dao/dao_factory.es6.js
```diff
@@ -0,0 +1,57 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+'use strict';
+
+foam.INTERFACE({
```
lukebjerring: It surprises me that passing in a DAO's class or type is not sufficient.

Does FOAM have a dependency injection framework, or equivalent?

lib/dao/worker_dao.es6.js
```diff
@@ -0,0 +1,74 @@
+// Copyright 2018 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+'use strict';
+
+require('./indexed_dao.es6.js');
+require('./dao_factory.es6.js');
+
+foam.CLASS({
+  name: 'WorkerDAO',
```
lukebjerring: Do we use this in more than one place? Feels like a lot of bloat just to fetch + cache a URL?

lib/dao/dao_factory.es6.js
mdittmer: Yes, we could use a class. My intention here was to make the configuration of underlying classes more explicit. E.g., we could have something like:

```js
foam.CLASS({
  name: 'CompatRestDAO',
  extends: 'foam.dao.RestDAO',

  properties: [
    {name: 'baseURL', value: 'https://web-confluence.appspot.com/compatDAO'},
  ],
});

.CLASS({
  name: 'WebInterfaceRestDAO',
  extends: 'foam.dao.RestDAO',

  properties: [
    {name: 'baseURL', value: 'https://web-confluence.appspot.com/webInterfaceDAO'},
  ],
});
/* ... */
```

I thought appropriately configured `RestDAOFactory` instances made more sense, but I don't have a strong feeling. we can just as easily use `remotDAOCls` and adopt the above pattern. Do you think using classes instead of factories is simpler/easier to understand/(better in some way I haven't thought of)?

lib/dao/worker_dao.es6.js
mdittmer: We use is multiple times in the same place. The hope is that the `api_service` will become something like:

```js
angular.service(/*.../ function() {
  return {
    releaseDAO: RestAndWorkerDAO.create({baseURLPath: `/releaseDAO`}),
    compatDAO: RestAndWorkerDAO.create({baseURLPath: `/compatDAO`}),
});
```

where `RestAndWorkerDAO` is a (not so great) name for the proxy that delegates directly to the REST API until the cached-in-worker DAO is ready. My main reason for "all the abstraction" here is to facilitate tests of each class with minimal baked in dependencies on code that isn't under test. This followed from me trying to think about how to make [this](https://github.com/mdittmer/confluence/blob/compat-catalog0/main/app2.es6.js) into testable parts of `api_service` + a UI that consumes the DAO.

I'm open to different approaches. Do you think all this wiring should just get baked in to `api_service` and we should write some kind of end-to-end test for `api_service` that somehow captures the network + caching behaviour?

mdittmer: @lukebjerring PTAL at latest push. It eliminates the extra factory abstraction.

PR #282, opened 2018-05-05
 Test and implement: New api_service DAO

mdittmer:
This encapsulates the logic for first-data + fast-(cached)-data on the main thread for `api_service`.

### DIFF
--- a/test/any/dao/api_service_dao-test.es6.js
+++ b/test/any/dao/api_service_dao-test.es6.js
@@ -149,6 +149,10 @@ describe('ApiServiceDAO', () => {
       workerRegistry: foam.box.Context.create(null, workerCtx).registry,
     }, ctx);
     const initialDelegate = dao.delegate;
+
+    // Assert that a property change event is fired by dao.delegate$ property
+    // slot. This test will time out (because done() isn't called) if
+    // dao.delegate is not set after this subscription is established.
     dao.delegate$.sub(done);
   });
 });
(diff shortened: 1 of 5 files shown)


test/any/dao/api_service_dao-test.es6.js
```diff
@@ -0,0 +1,154 @@
...
+    overwriteRegister(foam.dao.NullDAO, 'org.chromium.apis.web.WorkerDAO');
+    overwriteRegister(foam.dao.NullDAO, 'foam.dao.RestDAO');
+
+    const dao = pkg.ApiServiceDAO.create({
+      of: pkg.test.Thing,
+      baseURL: 'https://example.com/thingDAO',
+      workerRegistry: foam.box.Context.create(null, workerCtx).registry,
+    }, ctx);
+    const initialDelegate = dao.delegate;
+    dao.delegate$.sub(done);
```
lukebjerring: Am I missing something, or is there nothing being asserted in this test?

mdittmer: You're not missing anything. This test will time out if the `ApiServiceDAO` never changes its delegate after initialization. The code under test is the swapping out of the thread/process-local `RestDAO` for the now-ready `WorkerDAO`.

We could add assertions by reaching into the DAO and asserting that the delegate used to be a `WorkerDAO` and is now a `RestDAO`. I've been trying to keep the tests from depending too much on the implementation details though. WDYT? Should I add such assertions?

PR #283, opened 2018-05-07
Server-side changes for generated.CompatData DAO

mdittmer:
This exposes `/compatDAO` backed by `generated.CompatData` sourced in the usual way (i.e., same as `/releaseDAO`, `/webInterfaceDAO`, etc.). Because the code is generated, we also expose `/class:[pkg.path].generated.CompatData`, and also use it to bootstrap the `generated.CompatData` class in both front end and forked processes on the server.

### DIFF
--- a/lib/dao/json_dao_container.es6.js
+++ b/lib/dao/json_dao_container.es6.js
@@ -51,6 +51,7 @@ foam.CLASS({
     'org.chromium.apis.web.SerializableHttpJsonDAO',
     'org.chromium.apis.web.SerializableLocalJsonDAO',
     'org.chromium.apis.web.WebInterface',
+    'org.chromium.apis.web.generated.CompatData',
   ],
   imports: [
     'error',
@@ -89,6 +90,12 @@ foam.CLASS({
       },
     },
     {
+      name: 'compatDAO',
+      factory: function() {
+        return this.getDAO(this.CompatData);
+      },
+    },
+    {
       name: 'browserMetricsDAO',
       factory: function() {
         return this.getDAO(this.BrowserMetricData);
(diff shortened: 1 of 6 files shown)


lib/dao/json_dao_container.es6.js
```diff
@@ -116,6 +123,19 @@ foam.CLASS({
...
+      class: 'Function',
+      name: 'parseURL_',
+      transient: true,
+      factory: function() {
+        if (foam.isServer) {
+          const url = require('url');
+          return url.parse.bind(url);
+        } else {
+          return str => new URL(str);
+        }
```
lukebjerring: I'm sad about this code-smell - can you just verify that it's really needed?

main/forkScript.js
```diff
@@ -22,15 +23,31 @@ require('../lib/confluence/metric_computer_service.es6.js');
...
-  unsafe: false,
-  classWhitelist: require('../data/class_whitelist.json'),
-}, container));
+
+const compatClassFile = 'class:org.chromium.apis.web.generated.CompatData.json';
+
+// TODO(markdittmer): This should be local or remote based on param passed to
+// parent. It should be forwarded to forkScript invocation.
+const compatClassURL =
+    `${require('../data/http_json_dao_base_url.json')}/${compatClassFile}`;
```
lukebjerring: If it's just a string const, probably export its value as a sensible name, as the behaviour of the require/import.

lib/dao/json_dao_container.es6.js
mdittmer: The underlying issue appears to be that, in NodeJS, `URL` isn't in the global namespace. We could use `new require('url').URL(...)` (or equivalent) instead of `require('url').parse(...)`, but the reference to the URL library doesn't go away. An alternative approach would be to have some compat code that always runs early that sets `URL = typeof URL === 'undefined' ? require('url').URL : URL`. I'm not sure that's much better, though.

main/forkScript.js
mdittmer: Yeah. I think there are release process reasons this was isolated into a JSON file. I already have a TODO to change the way this line of code works anyway, so I'm going to leave it for now.

PR #285, opened 2018-05-08
Finish integrating components for new catalog UI

mdittmer:
This puts the `api_service` + UI code in a state that renders the new catalog UI :raised_hands: 

## Testing instructions

```bash
# Check out branch, ensure proper `npm install`...
npm run serve
# Wait for logging to slow/stop (10-30s depending on machine)
# Visit http://localhost:8080/#!/catalog
```

### DIFF
--- a/data/class_whitelist.json
+++ b/data/class_whitelist.json
@@ -1,4 +1,5 @@
 [
+  "FObject",
   "Property",
   "Requires",
   "String",
@@ -20,6 +21,10 @@
   "foam.box.SocketConnectBox",
   "foam.box.SubBox",
   "foam.box.SubBoxMessage",
+  "foam.core.FObject",
+  "foam.core.Property",
+  "foam.core.Requires",
+  "foam.core.String",
   "foam.dao.ArraySink",
   "foam.dao.ClientSink",
   "foam.dao.DAO",
--- a/lib/controller/api_catalog.es6.js
+++ b/lib/controller/api_catalog.es6.js
@@ -30,9 +30,9 @@ angular.module('confluence').controller('catalogController', ['api',
 
       const compatDAO = api.getApiServiceDAO(
           pkg.DAOContainer.COMPAT_NAME, CompatData);
-      const unorderdProperties = 
+      const unorderedProperties = 
           CompatData.getAxiomsByClass(pkg.CompatProperty);
-      const propDAO = foam.dao.ArrayDAO.create({array: unorderdProperties});
+      const propDAO = foam.dao.ArrayDAO.create({array: unorderedProperties});
       const E = foam.mlang.ExpressionsSingleton.create();
       const composeReleaseOrder = function() {
         let props = Array.from(arguments);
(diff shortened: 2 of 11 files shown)


data/class_whitelist.json
```diff
@@ -1,4 +1,5 @@
 [
+  "FObject",
```
lukebjerring: 2 `FObject`s?

lib/controller/api_catalog.es6.js
```diff
@@ -7,20 +7,86 @@ require('angular');
...
+    let CompatData;
     pkg.ClassGenerator.create({
       classURL: compatClassURL,
-    }).generateClass().then(CompatData => {
+    }).generateClass().then(Cls => {
+      CompatData = Cls;
+
       const compatDAO = api.getApiServiceDAO(
           pkg.DAOContainer.COMPAT_NAME, CompatData);
+      const unorderdProperties =
```
lukebjerring: nit: Typo (missing e)

lib/controller/api_catalog.es6.js
```diff
@@ -7,20 +7,86 @@ require('angular');
...
+      const orderedDAO = propDAO.orderBy(composeReleaseOrder(
+          pkg.Release.RELEASE_DATE,
+          pkg.Release.BROWSER_NAME,
+          pkg.Release.BROWSER_VERSION,
+          pkg.Release.OS_NAME,
+          pkg.Release.OS_VERSION));
+
+      return Promise.all([
+        orderedDAO.select(), 
+        orderedDAO.select(E.GROUP_BY(
```
lukebjerring: nit: Weird wrapping. better:

    orderedDAO.select(
      E.GROUP_BY(E.DOT(... ,
      foam.dao.ArraySink...

lib/controller/api_catalog.es6.js
```diff
@@ -7,20 +7,86 @@ require('angular');
...
+                E.DOT(pkg.CompatProperty.RELEASE, pkg.Release.BROWSER_NAME),
+            foam.dao.ArraySink.create())),
+      ]);
+    }).then(arraySinkAndGroupSink => {
+      const arraySink = arraySinkAndGroupSink[0];
+      const groupSink = arraySinkAndGroupSink[1];
```
lukebjerring: Unbox these badboys:

    const [arraySink, groupSink] = arraySinkAndGroupSink;

lib/controller/api_catalog.es6.js
```diff
@@ -7,20 +7,86 @@ require('angular');
...
+        selected,
+        removeColumn,
+      });
+  
+      const daoController = pkg.DAOController.create({
+        data: api.getApiServiceDAO(
+            pkg.DAOContainer.COMPAT_NAME, CompatData, ctx),
+      }, ctx);
+      const daoControllerView = pkg.DAOControllerView
+          .create(null, daoController);
```
lukebjerring: Can you name the null param

    .create(/*foo*/ null, daoController)

data/class_whitelist.json
mdittmer: I've learned that the default package resolution for whitelist contexts is sadly brittle:

Looking up `Foo` defaults to looking up `foam.core.Foo`. Fine. Reasonable. But the addition of the default package happens after the whitelist context performs its checks, so it's possible to have code `lookup('FObject')` in one place and `lookup('foam.core.FObject')` in another. They will resolve to the same class, but both need to be whitelisted :cry:

lib/controller/api_catalog.es6.js
mdittmer: Yeah. It's weird. I dropped it anyway. Just discovered a bug: order for default selected (by `browserName`) doesn't match order in browser selection list (which it should). I've implemented a different strategy that no longer uses DAO operations to cherry-pick the latest releases.

lib/controller/api_catalog.es6.js
mdittmer: No longer applicable with new strategy.

lib/controller/api_catalog.es6.js
mdittmer: I can add it, but it won't be very helpful IMHO. All FOAM classes implement `create(args, context)`; `create(/* args */ null, daoController)` doesn't say much about the meaning of the null because it's so generic.

lib/controller/api_catalog.es6.js
mdittmer: Done (push forthcoming).

data/class_whitelist.json
arobins: Maybe add a comment indicating why there are duplicates, so that a future editor has a chance of doing the thing that is required to make it work.

mdittmer: Sadly, I don't think comments are a part of the JSON spec. Maybe there should be a `data/README.md`?

PR #291, opened 2018-05-14
Port generated URLs from graphs to new catalog URL scheme

mdittmer:
URL scheme now has a collection of release and a single query that encodes keywords plus constraints like "API is/is-not in Release X". This change ports the generated URLs from interacting with metrics graph data points to this scheme.


lib/client/state.es6.js
```diff
@@ -79,18 +79,36 @@ foam.CLASS({
       of: 'org.chromium.apis.web.CompatProperty',
       documentation: 'Compat data properties associated with releases in view.',
       name: 'selected',
+      preSet: function(old, nu) {
+        if (!this.selectable) return nu;
+
+        const indexOf = (prop) => {
```
lukebjerring: const indexOf = prop => this.selectable.findIndex(sProp => sProp.name === prop.name)

lib/controller/api_confluence.es6.js
```diff
@@ -52,19 +53,41 @@ angular.module('confluence')
         $scope.aggressiveRemovalMetric = aggressiveRemovalMetric;
       });
 
-    //
-    // TODO(markdittmer): URL state needs to be updated according to new
-    // catalog.
-    //
+    const getTerseReleaseId = function(id) {
```
lukebjerring: This is a little hard to read. Add some tests.

Issue #201, opened 2017-09-26
Aliased webkit-prefixed CSS properties are missing in Chrome


RByers:
In Chrome, CSS properties which are aliases (alias_for in [CSSProperties.json5](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/css/CSSProperties.json5?q=CSSProperties.json5&sq=package:chromium&dr)) of other properties don't show up when enumerating the properties of a `CSSStyleDeclaration` instances, but they're actually still there:

```
> 'webkitUserSelect' in document.body.style
true
> Object.getOwnPropertyNames(document.body.style).indexOf('webkitUserSelect')
-1
```

This makes some CSS properties appear to be Safari-specific when they're really not.  I was thinking we should just fix this as a Chrome bug, but I see Safari has some like this as well - eg. `webkitFlex`.

Perhaps we should just have a hack for CSSStyleDeclaration to explicitly look for `'webkitFoo' in document.body.style`` for each `Foo` in `document.body.style`?

RByers:
I filed a [chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=768917) to discuss what should be done there, but I think that's probably orthogonal (best case, it'll be awhile before Chrome and Safari ship with these exposed, and there's enough properties impacted that I think hacking around it makes sense).

PR #294, opened 2018-05-15
Use object-graph-js BrowserObjectGraph

mdittmer:
Update use of `object-graph-js` to use `BrowserObjectGraph` class. Also, perform `npm update` to keep dependencies like `object-graph-js` and `foam2` fresh.


mdittmer:
This is towards #201

PR #296, opened 2018-05-18
ApiExtractor: Ignore only CONST_CASE constant values

mdittmer:
This further qualifies which object graph properties are ignored on the basis that they are constants. In particular, it will ignore based on the existing criteria + property-name-is-`CONST_CASE` criteria.

### DIFF
--- a/lib/web_catalog/api_extractor.es6.js
+++ b/lib/web_catalog/api_extractor.es6.js
@@ -685,6 +685,9 @@ foam.CLASS({
                   this.isConstName_(name));
     },
     function isConstName_(name) {
+      // A subset of the "identifier" expression in
+      // https://heycam.github.io/webidl/#idl-grammar
+      // that includes only identifiers like "FOO" and "FOO_BAR".
       return /^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$/.test(name);
     },
     function dedupConcat_(opt_arr1, arr2) {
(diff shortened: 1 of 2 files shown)


lib/web_catalog/api_extractor.es6.js
```diff
@@ -678,10 +678,14 @@ foam.CLASS({
...
-                // return true ignoring "value" metadata.
-                type === 'boolean'));
+                  (og.lookupMetaData(name, id).value === 1 ||
+                      // If type is boolean, value of false causes "value" to be
+                      // 0; return true ignoring "value" metadata.
+                      type === 'boolean') &&
+                  this.isConstName_(name));
+    },
+    function isConstName_(name) {
+      return /^[A-Z][A-Z0-9]*(_[A-Z0-9]+)*$/.test(name);
```
foolip: Can you link to https://heycam.github.io/webidl/#idl-grammar and say this is the all-caps subset of that? I think it doesn't matter too much if it's not an exact subset, leading underscore and dashes in the name is, I think, a hypothetical.

Could also simplify the expression somewhat to `/^[A-Z][A-Z0-9_]*$/` to allow for double underscores. Or comment here why *not* allowing double underscores is important.

test/any/web_catalog/api_extractor-integration.es6.js
```diff
@@ -92,10 +92,11 @@ describe('API extractor', function() {
           '+valueOf+': 6,
           'functionAPI': 10015,
           'InterfaceInstance': 10013,
-          'ObejectInstanceA': 10014,
+          'ObjectInstanceA': 10014,
           'property': 3,
           'constObjectProperty': 7,
-          'constantNumber': 3,
+          'constLikeNumber': 3,
+          'CONSTANT_NUMBER': 3,
```
foolip: Also test for double underscore if it remains part of the regex.

Issue #299, opened 2018-05-20
Filter out obvious non-APIs


mdittmer:
Interface names like `a` and `+toString+` currently show up on master branch data collection. @foolip are there any ligitimate interface names that would fail `name.length > 1 && /^[a-zA-Z]+$/.test(name)`?

foolip:
What are all the previously reported top-level things that would be excluded? Things with numbers come to mind, like Int8Array. As for short things, there isn't anything that disallows the introduction of single-char names, although I doubt it will ever happen. Where is `a` coming from? We have found accidentally leaked variable names from scripts and the like, and catching those is nice.

mdittmer:
Here's a summary of known non-APIs and where they come from:

- `a`: An old version of Safari contains a polyfill that mistakenly bled into the global object
- `+toString+`: This one is complicated. Basically, `+toString+` appears in the list of things on the global object because it's a leaky abstraction for `Object.prototype` member names, and `window` has `Object.prototype` in its prototype chain. Then, because our handling of "looks like a function" is currently, actually, "looks like a constructor" (checks for `.prototype` property), and `window.toString.hasOwnProperty('prototype')` is false, `window.toString` gets treated like a global library with the `name` API (from being a function).

I think we should have a `postProcessor` that filters out `a` and any other one-character interface names. I'm working on a multi-part fix for the `object-graph-js` leaky abstraction and differentiation between "looks like a function" and "looks like a constructor" in the data collection code.

foolip:
Fixed by https://github.com/GoogleChromeLabs/confluence/pull/301.

PR #300, opened 2018-05-23
 ApiExtractor: Detect functions and constructor functions differently

mdittmer:
Towards #299; handles `+toString+` interface name case.


mdittmer:
@foolip 

image

PR #305, opened 2018-06-18
Update metric names

### LINKED ISSUE #219: Rename the metrics to be as value-free as possible
See https://github.com/GoogleChrome/confluence/issues/178#issuecomment-333861976 an onward.

Ideas:
 - API Velocity → API count
 - Failure to Ship → Lone non-support? (Terrible, other ideas?)
 - Aggressive Removal → Lone removal
 - Browser-Specific: already OK?

To make it sounds more try, we could also use say "property" instead of API, which is technically accurate even for methods.

@bzbarsky @othermaciej

mdittmer:
Fixes #219


mdittmer:
It is now! https://update-metric-names-dot-web-confluence.appspot.com/

PR #316, opened 2018-08-01
Fonts and icons

mdittmer:
Lock-in a copy of font files from `materialize-css` and fix links from `materialize.css` (already copied and modified for custom CSS preferences).

Also, check in and refer to different sized icons for "modern favicon.ico" support. Followed instructions [here](https://sympli.io/blog/2017/02/15/heres-everything-you-need-to-know-about-favicons-in-2017/).


static/materialize.css
```diff
@@ -4396,31 +4396,31 @@ nav .input-field label.active i {
 
 @font-face {
   font-family: "Roboto";
-  src: local(Roboto Thin), url(/bundle/fonts/Roboto-Thin.woff2) format("woff2"), url(/bundle/fonts/Roboto-Thin.woff) format("woff");
+  src: local(Roboto Thin), url(/fonts/Roboto-Thin.woff2) format("woff2"), url(/fonts/Roboto-Thin.woff) format("woff");
```
arobins: The old path doesn't seem to exist, so I assume it was being set up as part of deployment. Should whatever config is causing that to happen be removed?

mdittmer: It was already removed when we dropped `require('materialize-css')` in favour of curating a modified CSS file. That change introduced a bug where the fonts that `materialize-css` was loading no longer got loaded. We could drop our webpack loader that knows how to deal with fonts, but I'm not sure much is gained from that. If we switch strategies later, it would be better for them to bundle correctly rather than break the build.

Make sense?

arobins: Sure.

Issue #338, opened 2019-01-25
Unit tests are broken


Hexcles:
`npm run test` currently fails on master.

foolip:
I found this too, and also found that it's failing on the r2018-11-01_00 branch. Here's the interesting part of the output I think:
```
..Assertion failed: Axiom name conflict in org.chromium.apis.web.CompatClass : alpha1_2_3zulu9_8_7
F..Assertion failed: Object graph JSON file name expected to be of the form:
                      window_[browser-name]_[browser-version]_[platform-name][platform-version].json
F

Failures:
1) CompatClassGenerator should fail console.assert() validating with duplicate relases
  Message:
    Expected function to throw an exception.
  Stack:
    Error: Expected function to throw an exception.
        at UserContext.it (/usr/local/google/home/foolip/src/confluence/test/node/web_apis/api_compat_data-test.es6.js:29:14)

2) ApiExtractorService should throw when "environment" data is missing and filename is malformed
  Message:
    Expected function to throw an exception.
  Stack:
    Error: Expected function to throw an exception.
        at UserContext.it (/usr/local/google/home/foolip/src/confluence/test/node/web_catalog/api_extractor_service-test.es6.js:58:58)

Ran 125 of 133 specs
125 specs, 2 failures
Finished in 6.926 seconds
```

It's not clear if the "Assertion failed" bits at the beginning are the cause of the failure or not, seems like maybe it *didn't* throw an exception and that's why the test failed?

@mdittmer is it obvious what's wrong here?

mdittmer:
It's not obvious but your read is correct: we are asserting that we hit an
error condition but we are not hitting the expected error condition. The
error output prior to the test report is likely expected, but perhaps
something that used to throw is logging? Or perhaps some work is being
deferred so that no exception is synchronously thrown? These are just
guesses. I'd have to step through the particular tests to discover the
cause.

foolip:
It looks like both the test and the code under test was were added in https://github.com/GoogleChromeLabs/confluence/pull/253.

Oddly, the tests don't seem to fail on Travis:
https://travis-ci.org/GoogleChromeLabs/confluence/builds/489954534

The confluence branch of foam2 isn't set up on Travis, perhaps there's a connection.

foolip:
It looks like the failures are specific to Node.js 10:
https://github.com/GoogleChromeLabs/confluence/pull/358

[...]

PR #343, opened 2019-02-12
Upgrade materialize-css to ^1.0.0

foolip:
This is to get rid of the jquery dependency that materialize-css had.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,10 +29,10 @@
       "integrity": "sha1-c7Xuyj+rZT49P5Qis0GtQiBdyWU=",
       "dev": true,
       "requires": {
-        "co": "4.6.0",
-        "fast-deep-equal": "1.1.0",
-        "fast-json-stable-stringify": "2.0.0",
-        "json-schema-traverse": "0.3.1"
+        "co": "^4.6.0",
+        "fast-deep-equal": "^1.0.0",
+        "fast-json-stable-stringify": "^2.0.0",
+        "json-schema-traverse": "^0.3.0"
       }
     },
     "align-text": {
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -11274,180 +11278,21 @@
...
-            "once": "1.4.0",
-            "wrappy": "1.0.2"
-          }
-        },
-        "inherits": {
-          "version": "2.0.3",
-          "resolved": "https://registry.npmjs.org/inherits/-/inherits-2.0.3.tgz",
-          "integrity": "sha1-Yzwsg+PaQqUC9SRmAiSA9CCCYd4="
-        },
-        "jquery": {
```
foolip: Dropping this is the important part.

package.json
```diff
@@ -33,7 +33,7 @@
     "d3": "^4.7.4",
     "foam2": "git://github.com/foam-framework/foam2.git#confluence",
     "jquery": "^3.2.1",
```
Hexcles: But we still directly depend on jquery here?

foolip: Yes, we depended on two of them! Grepping for "jquery" shows only config/webpack.common.js using it, but if it's doing something there I wonder if it could be a dev dependency. @mdittmer can you advise?

Hexcles: All it does is to provide jquery to the front-end, but I don't see anything actually using it. I suspect we can simply get rid of it.

foolip: I'll do that in a separate PR so we can bisect any regression more easily.

Hexcles:
I'm merging this despite the Travis failure (it's a known issue #338 ).

PR #344, opened 2019-02-12
Add changes from mdn branch in mdittmer/confluence

foolip:
https://github.com/mdittmer/mdn-confluence has a dependency on "git://github.com/mdittmer/confluence.git#mdn".

This PR brings in those changes to make it more straightforward to make changes. In particular, there are `npm audit` issues that bubble all the way to BCD that needs to be addressed.


foolip:
@mdittmer could you give this a quick look and call out any changes you know aren't needed for mdn-confluence in BCD?

foolip:
https://github.com/mdittmer/mdn-confluence/blob/master/package.json now depends on git://github.com/GoogleChromeLabs/confluence.git#mdn-confluence-2 instead.

PR #345, opened 2019-02-14
Fix one failing unit test

foolip:
At one point `foam.assert` must have thrown an exception, but is now
just a wrapper for `console.assert` and does not throw. Replace this
with conditional `throw new Error` which is already used elsewhere.

Part of https://github.com/GoogleChromeLabs/confluence/issues/338.


foolip:
That's weird, it passed for me locally but not on Travis. I'll try reverting to fix just the one test.

foolip:
The remaining failure seems to be because I put a slash in the branch name :)

PR #346, opened 2019-02-15
Remove the jquery dependency

foolip:
With this change no further jquery remains in package-lock.json.

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -58,10 +58,6 @@ module.exports = {
   },
   plugins: [
     new CleanWebpackPlugin([C.BUNDLE_PROJECT_DIR], {root: C.ROOT_DIR}),
-    new webpack.ProvidePlugin({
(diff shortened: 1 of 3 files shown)


Hexcles:
The staging deployment https://rm-jquery-dep-dot-web-confluence.appspot.com looks correct to me. I'm merging this PR.

foolip:
Cool, having staging deployment working again makes things easier :)

foolip:
Turns out this did break stuff, the broken graphs I've mentioned in https://github.com/GoogleChromeLabs/confluence/issues/350#issuecomment-483144079 seem to be because of a jQuery dependency. Trying to sever it now.

Hexcles:
Sigh, having to ship jQuery is tremendously unfortunate.

foolip:
@Hexcles I was able to avoid it in the end :)

PR #347, opened 2019-02-18
Remove ALL dependencies (none needed for use with mdn-confluence)

foolip:


foolip:
I've confirmed that this works and it will significantly shrink the dependency tree for BCD. This can't be merged before https://github.com/mdittmer/mdn-confluence/pull/41 however, because scripts/build.js depends on the foam2 dependency existing.

foolip:
I think the only way to not break already published versions of mdn-confluence is actually going to be going with a new branch. I'll push the same contents to mdn-confluence-2 and close this PR.

PR #348, opened 2019-02-20
Upgrade css-loader to ^2.1.0

foolip:
This is to fix https://nodesecurity.io/advisories/654.


foolip:
https://css-loader-dot-web-confluence.appspot.com seems like it works, but I'm not sure what breakage would look like if it didn't. @mdittmer can you help review this?

PR #352, opened 2019-04-18
Remove trailing whitespace

foolip:


foolip:
Produced as a dummy commit to try deploy with *some* local change, might as well submit it.

PR #353, opened 2019-04-18
Rewrite stray jQuery dependency used for Materialize Tabs

foolip:
Broken by https://github.com/GoogleChromeLabs/confluence/pull/346 and
noticed when trying to deploy from scratch. In previous efforts to
verify the webpack was probably not updated.

The UI component used here is https://materializecss.com/tabs.html


foolip:
TBRing so I can deploy and go home :)

Hexcles:
Great! We don't need jQuery then :)

PR #354, opened 2019-04-18
Allow deploying without symlinking node_modules/foam2

foolip:
The exact version of foam2 deployed is saved in package-lock.json and
not commiting any change results in "Your working tree is not clean",
so since we don't need to change the version of foam2 used without
also updating the confluence repo, deploy can be simplified.

This is the only (apparent) remaining reason that node_modules/foam2
needs to be symlinked to a full foam2 checkout.

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -25,35 +25,15 @@ CONFLUENCE_CLEAN_OUTPUT=$(git clean --dry-run)
 CONFLUENCE_VERSION=$(git rev-parse HEAD)
 popd > /dev/null
 
-pushd "${WD}/../node_modules/foam2" > /dev/null
-FOAM2_DIFF_OUTPUT=$(git diff --stat)
-FOAM2_CLEAN_OUTPUT=$(git clean --dry-run)
-FOAM2_BRANCH=$(git rev-parse --abbrev-ref HEAD)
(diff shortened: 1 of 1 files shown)


foolip:
https://confluence-6b059be-dot-web-confluence.appspot.com/ was deployed using `npm run deploy` and the ongoing Travis job should also deploy soon. If both work then this worked :)

Hexcles:
If version is already locked by `package-lock.json`, why did we need this step in the first place? Perhaps @mdittmer could shed some lights?

On the other hands, I think `package-lock.json` will be automatically updated if you run `npm up`. I think we need to lock the version explicitly in `package.json` (the default is probably a range) instead of `package-lock.json`.

Hexcles:
Never mind, this was already done:

https://github.com/GoogleChromeLabs/confluence/blob/5c1fea61157b2dc28e8d115b6dc42712d0a5480b/package.json#L34

foolip:
https://simplify-deploy-dot-web-confluence.appspot.com/ also worked, although the commenting is broken so I had to find it in the logs.

foolip:
@Hexcles that branch could change so it's not exactly locked, but I doubt it will ever be hard to figure out which version was used given how little we change it.

foolip:
(Other packages aren't locked to an exact version either.)

Issue #328, opened 2018-11-12
Download broken?


gsnedders:
From https://web-confluence.appspot.com/#!/catalog?releases=%5B%22Safari_11.1_OSX_10.13.4%22,%22Edge_17.17134_Windows_10.0%22,%22Chrome_70.0.3538.67_Windows_10.0%22,%22Firefox_62.0_Windows_10.0%22%5D&q=%22in:saf11.1%20in:chr70%20notin:edg17%20notin:fir62%22 hitting the download button outputs a 207 byte file which is clearly incomplete.

foolip:
I've discovered this as well. Need to either fix or remove the download button.

foolip:
Here's the code that's misbehaving, it's client side JS:
https://github.com/GoogleChromeLabs/confluence/blob/16e1e21e21496e261db2165631fbdf40a723780d/lib/u2/DAOControllerView.es6.js#L293-L320

But the last change to that file was the commit (16e1e21e21496e261db2165631fbdf40a723780d) that introduced the functionality, so the regression must be found elsewhere.

foolip:
I'm pretty sure this is an escaping problem and https://crbug.com/123004 specifically, so a change in Chrome and not Confluence. @stephenmcgruer it may amuse you to learn it affected me, but I'm no less enthusiastic about the change :)

foolip:
Fixed in https://web-confluence.appspot.com/#!/ now.

gsnedders:
Thanks for finally fixing this @foolip!

PR #355, opened 2019-04-23
Encode data: URL correctly in download handler

foolip:
Fixes https://github.com/GoogleChromeLabs/confluence/issues/328.


foolip:
Confirmed working with https://confluence-bcbdd2d-dot-web-confluence.appspot.com.

PR #360, opened 2019-04-26
Test both Node.js 8 and 10 in CI

foolip:


foolip:
The second commit fixes https://github.com/GoogleChromeLabs/confluence/issues/338.

PR #363, opened 2019-04-27
Revert back to `runtime: nodejs` for app.yaml, meaning Node.JS 10

foolip:
Changed to nodejs8 in https://github.com/GoogleChromeLabs/confluence/pull/360,
but it doesn't actually work. The version of Node.JS used in AppEngine flex is
determined by the Docker base image, and is currently 10.15.3:
https://github.com/GoogleCloudPlatform/nodejs-docker/pull/196

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -2,7 +2,7 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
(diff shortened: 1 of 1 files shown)


foolip:
Followup is https://github.com/GoogleChromeLabs/confluence/pull/364

PR #367, opened 2019-04-30
Enable lint rules that can be fixed automatically

foolip:
Code changes are entirely from `npm run lint -- --fix`.


foolip:
Cool, the code coverage setup seems to actually work. Coverage has decreased because some reformatting resulted in more lines :)

PR #369, opened 2019-05-03
Log and ignore promise rejection in ApiServiceDAO delegate

foolip:
Because this turns into an unhandled promise rejection, after
upgrading to Jasmine 3, it would cause some subsequent tests to fail
in a mysterious manner. (Pinpointing the cause was not easy.)

Note that this would have been caught by this lint rule:
https://github.com/xjamundx/eslint-plugin-promise/blob/master/docs/rules/catch-or-return.md

### DIFF
--- a/lib/dao/api_service_dao.es6.js
+++ b/lib/dao/api_service_dao.es6.js
@@ -22,6 +22,8 @@ foam.CLASS({
     'foam.dao.RestDAO',
     'org.chromium.apis.web.WorkerDAO',
   ],
+  
+  import: ['warn'],
 
   properties: [
     {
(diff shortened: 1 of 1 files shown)


lib/dao/api_service_dao.es6.js
```diff
@@ -62,6 +62,8 @@ foam.CLASS({
         // Change delegate to client once it has finished caching in worker.
         clientDAO.limit(1).select().then(() => {
           this.delegate = clientDAO;
+        }, (err) => {
+          this.warn(`Caching all ${this.of.id} data failed. Using REST API instead. Reason: ${err}`);
```
mdittmer: This class should `import: ['warn']`. If it's already working, then it is by an accident of inheritance that we shouldn't rely on.

foolip: Alright, I removed it after a warning about another class, but that didn't remove the warning so I'll fix that separately.

PR #371, opened 2019-05-09
Remove unused variables and enable lint rule for it

foolip:


foolip:
The only interesting change is in test/any/dao/api_service_dao-test.es6.js

PR #376, opened 2019-05-09
Replace istanbul with nyc

foolip:
Rather than trying to migrate the `istanbul-combine` and
`istanbul check-coverage` bits, just rely on `codecov` to merge
reports, and have no checking/reporting of coverage locally.

This fixes most `npm audit` issues.


foolip:
I split this out from a bigger change also include jasmine to show that the apparent small regression in coverage has to do with changes in instrumentation rather than the test framework breaking. Looking at the coverage diff, it's mostly the first line of function definitions (`function foo() {`) that have changed, things like this:
https://codecov.io/gh/GoogleChromeLabs/confluence/pull/376/src/lib/web_catalog/object_graph_importer.es6.js

PR #377, opened 2019-05-09
Upgrade jasmine and karma

foolip:
Jasmine 3.0 defaults to running tests in random order
(https://github.com/jasmine/jasmine/blob/master/release_notes/3.0.md)
and this breaks api_count-test.es6.js, so disable random order tests
to keep the old behavior. (Might be enabled separately.)

Previous changes that would have blocked this:
 - https://github.com/GoogleChromeLabs/confluence/pull/366
 - https://github.com/GoogleChromeLabs/confluence/pull/369

---

Based on https://github.com/GoogleChromeLabs/confluence/pull/376 (first commit) and needs to be rebased after that has been merged.


foolip:
https://codecov.io/gh/GoogleChromeLabs/confluence/compare/istanbul-to-nyc...jasmine-karma shows no change in coverage, which gives some confidence that the tests are still running correctly. Something *might* still have silently broken, but if so that thing didn't have any unique code coverage.

PR #392, opened 2019-05-12
Upgrade webpack and related packages

foolip:
The `mode` option needs to be added to webpack config:
https://webpack.js.org/migrate/4#mode

CommonsChunkPlugin is replaced by new config:
https://webpack.js.org/migrate/4#commonschunkplugin
https://webpack.js.org/configuration/optimization/#optimizationsplitchunks

CleanWebpackPlugin uses webpack's `output.path` by default, so use the
default configuration:
https://github.com/johnagan/clean-webpack-plugin/issues/106

UglifyjsWebpackPlugin is replaced with TerserWebpackPlugin, which is
now webpack's default minimizer:
https://webpack.js.org/plugins/terser-webpack-plugin/#terseroptions
https://github.com/webpack/webpack/pull/8036

Unfortunately karma-webpack 4.0.0 isn't released yet, so a release
candidate is used instead.

### DIFF
--- a/config/webpack.common.js
+++ b/config/webpack.common.js
@@ -5,7 +5,6 @@
 
 const path = require('path');
 
-const webpack = require('webpack');
 const CleanWebpackPlugin = require('clean-webpack-plugin');
 
 const C = require('./webpack.constants.js');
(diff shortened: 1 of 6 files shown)


foolip:
The total bundle size is made slightly smaller by this:
```
$ ls -sh old-static/bundle/ static/bundle/
old-static/bundle/:
total 2.7M
104K app.bundle.js  692K foam.bundle.js  1.9M vendors.bundle.js   44K worker.es6.bundle.js

static/bundle/:
total 2.3M
104K app.bundle.js  696K foam.bundle.js  1.5M vendors.bundle.js   44K worker.es6.bundle.js
```

I suspect that a lot of vendors.bundle.js is unused code, but I won't change that in this PR.

config/webpack.common.js
```diff
@@ -56,14 +55,19 @@ module.exports = {
...
     ],
   },
-  plugins: [
-    new CleanWebpackPlugin([C.BUNDLE_PROJECT_DIR], {root: C.ROOT_DIR}),
-    new webpack.optimize.CommonsChunkPlugin({
-      name: 'vendors',
-      minChunks: function(module) {
-        return module.context && module.context.indexOf('node_modules') !== -1;
+  optimization: {
+    splitChunks: {
```
foolip: This is from https://webpack.js.org/plugins/split-chunks-plugin/#split-chunks-example-2 and I think it's equivalent to the original config.

foolip:
Staging deployment: https://confluence-8acbeef-dot-web-confluence.appspot.com

foolip:
Tested manually, seems fine.

foolip:
Neither am I, the commit message is all that I learned trying to fix this :)

PR #395, opened 2019-05-13
Upgrade to Babel 7

foolip:
https://babeljs.io/docs/en/v7-migration

There's no need to run babel at all when running tests or with a local
dev server, so drop `ES6_LOADER_OPTIONS_DEV` entirely.

The use of `comments: false` in the babel options makes no difference
at all to the size of the bundles, presumably because comments are
stripped again by the minimizer, so that option is not preserved.

---

This PR build on https://github.com/GoogleChromeLabs/confluence/pull/392 and needs to be rebased after that is merged. The "Upgrade webpack and related packages" commit should not be reviewed here.


config/webpack.constants.js
```diff
@@ -14,41 +14,6 @@ const FOAM_BIN_TMP_DIR = `${ROOT_DIR}/.local`;
...
-    'transform-es2015-block-scoping',
-    'transform-es2015-arrow-functions',
-    'transform-es2015-template-literals',
-    ['transform-es2015-modules-commonjs', {strict: false}],
-  ],
-  // No, really. Without strict mode.
-  parserOpts: {strictMode: false},
-};
-const ES6_LOADER_OPTIONS_DEV = {
-  plugins: ES6_LOADER_OPTIONS_PROD.plugins.concat(['transform-runtime']),
```
foolip: The code coverage has changed and it comes down mostly this line. On master, just commenting out this line but still using babel causes the number of lines to decrease. Not using babel at all (as in this PR) decreases the number functions/etc. too.

I suspect that because of the transformation some of the coverage information will have been bogus, but will have to check a bit more.

foolip: Addressed by landing https://github.com/GoogleChromeLabs/confluence/pull/398 first.

foolip:
The codecov comment is still comparing to an old commit on master, but https://codecov.io/gh/GoogleChromeLabs/confluence/compare/master...babel7 now shows that there's no coverage regression for the single commit on this PR.

PR #397, opened 2019-05-14
Enable the no-undef lint rule

foolip:
This did reveal a real mistake in main/forkScript.js, although in code
that doesn't really matter if it's broken because it would
`process.exit(1)` right afterwards anyway.

### DIFF
--- a/lib/controller/api_confluence.es6.js
+++ b/lib/controller/api_confluence.es6.js
@@ -3,6 +3,8 @@
 // found in the LICENSE file.
 'use strict';
 
+/* global M */
+
 require('angular');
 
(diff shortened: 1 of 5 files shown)


lib/controller/api_confluence.es6.js
```diff
@@ -3,6 +3,8 @@
 // found in the LICENSE file.
 'use strict';
 
+/* global M */
```
Hexcles: Perhaps briefly mention what is `M` (I suppose it's an Angular JS thing? But the keyword is too short to search for.)

foolip: Added a comment: M is Materialize

PR #398, opened 2019-05-14
Don't use babel when running tests using karma

foolip:
Currently instrumentation is running after babel, which results in
incorrect and nonsensical coverage information. The node vs. web
coverage here is an example:
https://codecov.io/gh/GoogleChromeLabs/confluence/src/f6802df5a6be89042a1db1e7b82f0a0a90906caf/lib/dao/indexed_dao.es6.js

This could be fixed by changing the order:
https://github.com/GoogleChromeLabs/confluence/pull/404

However, it's not clear if that's correct:
https://github.com/webpack-contrib/istanbul-instrumenter-loader/issues/86

There's no need to run Babel to test stable versions of Chrome or Firefox,
so just skip babel-loader.

### DIFF
--- a/config/karma.conf.js
+++ b/config/karma.conf.js
@@ -108,15 +108,6 @@ configurator.webpackConfig = {
   module: {
     rules: [
       {
-        test: C.ES6_REG_EXP,
-        use: [
-          {
-            loader: 'babel-loader',
-            options: C.ES6_LOADER_OPTIONS_DEV,
-          },
-        ],
-      },
-      {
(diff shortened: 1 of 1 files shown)


foolip:
https://github.com/webpack-contrib/istanbul-instrumenter-loader/issues/86 is an issue about this that I think is related to the problem.

foolip:
The change in coverage is of course because of the semi-random offsets of line coverage before, which would incorrectly increase line coverage.

PR #402, opened 2019-05-15
Split Travis into one job per coverage group

foolip:


foolip:
@Hexcles this builds on https://github.com/GoogleChromeLabs/confluence/pull/401 which I self-reviewed.

foolip:
The lack of codecov comment on this PR is concerning. The push build finished before I created the PR, but maybe I've also broken codecov integration. Forcing an update to see what will happen.

foolip:
Alright, seems fine, was probably a codecov infra issue.

.travis.yml
```diff
@@ -1,16 +1,29 @@
...
+    - npm run coverageNode
+
+  - name: firefox
+    node_js: 10
+    addons:
+      firefox: latest
+    before_script:
+    - export DISPLAY=:99.0
+    - sh -e /etc/init.d/xvfb start
+    - sleep 3 # give xvfb some time to start
```
Hexcles: This makes me a bit uneasy, but I see this is the recommended way on Trusty (https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-xvfb-directly).

foolip: It is pretty unsightly, using xvfb-run might be an alternative, but then I'd rather look into running headless Chrome and Firefox somehow. So I'll leave it be :)

foolip: Gave it a try: https://github.com/GoogleChromeLabs/confluence/pull/405

PR #403, opened 2019-05-15
Try `enforce: 'post'` for istanbul-instrumenter-loader

foolip:
https://github.com/webpack-contrib/istanbul-instrumenter-loader#with-babel


foolip:
This was an experiment for https://github.com/GoogleChromeLabs/confluence/pull/398.

PR #405, opened 2019-05-15
Run Karma with headless Chrome and Firefox

foolip:


foolip:
I verified it works locally. Will later build on this to test Chrome in Travis as well.

config/karma.conf.js
```diff
@@ -77,7 +77,7 @@ function configurator(config) {
     // available browser launchers:
     // https://npmjs.org/browse/keyword/karma-launcher
     // If run in travisCI, use firefox, otherwise, use Chrome.
```
Hexcles: I didn't notice this before. Why do we use different browsers in different envs? Can we also have Chrome running on Travis?

foolip: I guess it was to make it faster originally. I'll send a PR after this to also run Chrome on Travis.

foolip:
https://codecov.io/github/GoogleChromeLabs/confluence/commit/7419116cd9557352efdc6b7a41fea214c56980ca looks good.

PR #407, opened 2019-05-16
Run web tests in both Chrome and Firefox on Travis

foolip:
This also runs in both locally by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,19 @@ matrix:
     - npm run lint
     - npm run coverageNode
 
+  - name: chrome
+    node_js: 10
+    addons:
+      chrome: stable
+    script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -18,12 +18,19 @@ matrix:
     - npm run lint
     - npm run coverageNode
 
+  - name: chrome
+    node_js: 10
+    addons:
+      chrome: stable
```
Hexcles: Funny how Travis decided to use "stable" for Chrome and "latest" (which I believe is also stable) for Firefox.

foolip: Yep, I had to read https://docs.travis-ci.com/user/firefox to understand that latest means stable, and it looks like it's called that to match Firefox URL patterns.

PR #408, opened 2019-05-16
Fix and enable data quality integration tests

foolip:


foolip:
I consider this blocked on creating much smaller test data and putting those in test/data/.

foolip:
I'm never going to finish this.

PR #409, opened 2019-05-16
Use the URL constructor instead of url module

foolip:
The URL class is available on the global object since nodejs 10:
https://nodejs.org/api/url.html#url_the_whatwg_url_api


foolip:
There was a regression: https://github.com/GoogleChromeLabs/confluence/pull/411

foolip:
Reverted in https://github.com/GoogleChromeLabs/confluence/pull/412 instead.

PR #411, opened 2019-05-17
Don't use URL constructor for protocol-less file system paths

foolip:
This regressed with https://github.com/GoogleChromeLabs/confluence/pull/409.

The code path isn't tested, but is used by local_og_to_confluence.sh.

The problem is that `require('url').parse('/a/b/c.json')` returns a
URL with `protocol` null and pathame '/a/b/c.json', while the URL
constructor throws since it's an invalid URL.

Fixing this by prepending `file:` wouldn't even normalize the
pathname, so just don't use the URL constructor.


foolip:
Reverted in https://github.com/GoogleChromeLabs/confluence/pull/412 instead.

PR #412, opened 2019-05-17
Revert "Use the URL constructor instead of url module"

foolip:
This reverts https://github.com/GoogleChromeLabs/confluence/pull/409.


foolip:
This broke more than what I fixed in https://github.com/GoogleChromeLabs/confluence/pull/411. `npm run serve` was also broken because of something to do with URLs. To make the change the code paths would have to be vetted more, so I'm reverting.

PR #417, opened 2019-05-22
Add Firefox 67 release date

foolip:
Matches https://github.com/mdn/browser-compat-data/pull/4149


foolip:
Data in https://github.com/mdittmer/web-apis/pull/84

foolip:
Thanks @mdittmer!

PR #469, opened 2019-08-20
Filter out `__REACT_DEVTOOLS_GLOBAL_HOOK__`

foolip:
Also fix a URL broken by https://github.com/GoogleChromeLabs/confluence/pull/339.

### DIFF
--- a/lib/web_catalog/api_extractor.es6.js
+++ b/lib/web_catalog/api_extractor.es6.js
@@ -256,6 +256,10 @@ foam.CLASS({
             interfaceNames: this.blacklistInterfaces,
           }),
           this.RemoveApisProcessor.create({
+            interfaceNames: ['Window'],
+            apiNameRegExp: /^__REACT_DEVTOOLS_GLOBAL_HOOK__$/,
+          }),
+          this.RemoveApisProcessor.create({
             interfaceNames: ['CSSStyleDeclaration'],
             apiNameRegExp: /[-]/,
             confluenceIssueURL: '',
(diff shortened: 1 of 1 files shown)


lib/web_catalog/api_extractor.es6.js
```diff
@@ -209,6 +209,11 @@ foam.CLASS({
           'window',
           // Accidentally shipped in Safaris 5 - 8.
           'a',
+          // React DevTools extension installed on BrowserStack
```
mdittmer: nit: Period at end of comment.

lib/web_catalog/api_extractor.es6.js
```diff
@@ -247,10 +252,16 @@ foam.CLASS({
           this.RemoveInterfacesProcessor.create({
             otherURLs: [
               // For interface name "a":
-              'https://github.com/GoogleChromeLabsLabs/confluence/issues/299#issuecomment-391119693',
+              'https://github.com/GoogleChromeLabs/confluence/issues/299#issuecomment-391119693',
+              // For interface name "__REACT_DEVTOOLS_GLOBAL_HOOK__":
```
mdittmer: I think this belongs on the newly added `RemoveApisProcessor` below; setting [this property](https://github.com/GoogleChromeLabs/confluence/blob/578b00eae6a0b04409b7360ddb2409fb30704ef6/lib/web_catalog/post_processors.es6.js#L37-L42) on the appropriate instance. (Should be, as here, an array with one element.)

foolip:
Thanks for the review, @mdittmer! I finally got around to addressing the feedback as I'd like to make a new release. I'll merge this now.

PR #504, opened 2019-11-15
Make an async test really async (was broken)

foolip:
This was caught by eslint-plugin-promise, but that config isn't being
added at this point.


foolip:
@stephenmcgruer can you review?

foolip:
If this test had failed, it would have turned into an unhandled promise rejection.

PR #567, opened 2020-03-27
Increase maxAssetSize and maxEntrypointSize

Hexcles:
With the recent updates, our 3rd-party dependencies have grown over
1.5MiB, so we need to adjust the maxAssetSize to suppress an error;
maxEntrypointSize also needs to be adjusted as it includes vendors, too.

TBR @foolip to unblock release

### DIFF
--- a/config/webpack.prod.js
+++ b/config/webpack.prod.js
@@ -49,7 +49,7 @@ module.exports = merge(common, {
     // recommendation of 250kb, but are combined with 'error' to ensure that
     // they don't grow accidentally beyond the current size.
     hints: 'error',
-    maxAssetSize: 1.5*1024*1024,
-    maxEntrypointSize: 2.2*1024*1024,
+    maxAssetSize: 2*1024*1024,
+    maxEntrypointSize: 2.5*1024*1024,
   },
 });


stephenmcgruer:
Just increasing the error size doesn't seem like a great solution here. Normally webpack tracks like 250kb as a goal, and we're almost at an order of magnitude higher?! Can you open an issue to investigate this, if it isn't something already well known?

Hexcles:
I picked a few versions dating way back to Nov 1, 2018 (the last release by Mark). I'm seeing a steady increase of the vendor bundle size over time (about 30% gradually in the past two years). I think this bundle probably needs to be split into multiple chunks, but the front-end hard codes the asset URL. 

All in all, this vendor bundle is still only ~400KiB when gzipped and this project is currently in maintenance mode. I might come back and revisit perhaps after fixing the more pressing performance issue of wpt.fyi...

stephenmcgruer:
> All in all, this vendor bundle is still only ~400KiB when gzipped and this project is currently in maintenance mode. I might come back and revisit perhaps after fixing the more pressing performance issue of wpt.fyi...

I am not suggesting we fix it now, I am suggesting we don't forget about it :). Filed https://github.com/GoogleChromeLabs/confluence/issues/568, just as a reminder to future-us.

PR #679, opened 2020-11-02
Require Node.js 12

foolip:


foolip:
The result of this was "Installed Node.js v15.0.1". Don't know what version was previously used.

foolip:
Without this change, one gets "Your package.json does not specify a supported Node.js version.  Please pin your application to a major version of the Node.js runtime.  To learn more, visit https://cloud.google.com/appengine/docs/flexible/nodejs/runtime" I can't tell what's actually installed.

